PR #11, opened 2019-11-14
Use `zip::zipr()` instead of `zip::zip()`

davidgohel:
Thanks for maintaining that great package. 

This PR will make disappear the note about deprecated `zip:zip`. I Hope you will find it useful.

David


ycphs:
Thank you

Issue #13, opened 2019-11-19
Writing NA's as a custom string


soliac:
## Desired behaviour
When writing confidentialised data, with suppressed rows encoded as NA's, I want to fill each NA cell with an "S" character.

## Actual Behavior
Because this mixes numeric and character types, I cannot create this data in R and write the whole column out while preserving the different types. Instead, I have to write out an "S" to each suppressed cell individually.

## Proposed solution
Add an "na.string" parameter that defaults to NULL. If "keepNA" is TRUE, and "na.string" is not NULL, write na.string as a string to any cells containing NA's.

soliac:
Proposed code change in pull request #14

ycphs:
Thank you creating the pull request. I already merged it. I think the issue is now ready to be closed.

PR #14, opened 2019-11-19
Write NA's as na.string

soliac:
Add an "na.string" parameter that defaults to NULL. If "keepNA" is TRUE, and "na.string" is not NULL, write na.string as a string to any cells containing NA's.


soliac:
Related feature request #13

ycphs:
Thank you.

PR #63, opened 2020-04-20
Write mixed EDT/EST DateTimes

blakeboswell:
This PR adapts the existing write to worksheet code so that all values' timezones in a datetime column are considered rather than only the first non-na value's timezone.

fixes issue #64

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 *  Fixed NEWS hyperlink
 
+*  Fixed writing of mixed EST/EDT datetimes.
+
 # openxlsx 4.1.4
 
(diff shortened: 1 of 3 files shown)


blakeboswell:
passes `devtools::check()` on 

```
─  using R version 3.6.3 (2020-02-29)
─  using platform: x86_64-apple-darwin15.6.0 (64-bit)
```

will try and figure out why it's failing here before modifying news.  any ideas appreciated.

blakeboswell:
not sure what's causing the issue ... seems ok here:
https://travis-ci.org/github/blakeboswell/openxlsx

PR #67, opened 2020-05-06
Styler

ycphs:
added style_pkg to the package 
with cran release prep


ycphs:
added to the master
cran release prep

PR #85, opened 2020-07-24
New Feature: group/ungroup columns/rows

### LINKED ISSUE #32: Feature Request: Add Group and Ungroup columns & rows
Hi @ycphs 

I have a problem to output a data frame to xlsx with certain styling like grouping several columns based on some conditions.

I think it will be great if openxlsx can have functionality to do grouping & ungrouping columns & rows, while user need to do some manipulation on data frame.

Currently I am doing it in a VBscript, and call it from R, which hard to trace bugs.

image
Here is the functionality as discussed, under 'Data' tab in excel

JoshuaSturm:
Added the functions `groupColumns()` and `groupRows()`, as well as their counterparts `ungroupColumns()` and `ungroupRows()`.

Note that both `groupColumns()` and `setColWidths()` have a `hidden` parameter; modifying one will update the other (if a conflict exists).

Resolves #32.


ycphs:
Thank you for the great work.

ycphs:
I just removed the warning "warning: may be used uninitialized in this function" with the last [commit ](https://github.com/ycphs/openxlsx/commit/b0dd363f8c8e5c0165cc375cd18d624859143ecf)

ycphs:
The new feature seems to have an issue: #92 

Could you please have a deeper look into it?

PR #91, opened 2020-08-14
Make baseColWidth use Excel's default

tobwen:
Setting `baseColWidth` to `10` synchronizes the column width to Microsoft's XLSX standard since Excel 2007.


tobwen:
More details: https://github.com/awalker89/openxlsx/issues/425

ycphs:
thank you very much

PR #104, opened 2020-09-27
Fix index for `hidden` attribute when grouping columns

### LINKED ISSUE #100: Error when using setColWidths() and groupColumns() on overlapping columns
When using both functions `setColWidths()` and `groupColumns()` on overlapping columns I get the following error:
```R
Error in if (width_hidden != outline_hidden) { : 
  missing value where TRUE/FALSE needed
```

**To Reproduce**
```R
rm(list = ls())
library(openxlsx)

# create workbook and add sheet
wb <- createWorkbook(title = "column width and grouping error")
addWorksheet(wb, sheetName = 1)

# set col widths
setColWidths(
  wb,
  sheet = 1,
  cols = 1:100,
  widths = 8
)

# group columns
groupColumns(wb, sheet = 1, cols = 20:100, hidden = TRUE)
```

results in
```R
Error in if (width_hidden != outline_hidden) { : 
  missing value where TRUE/FALSE needed
```

**Expected behavior**
I expected that groupColumns just would group it afterwards, but clearly there is something wrong with the behaviour of the function.

**Desktop :**
```R
> sessionInfo()
R version 3.6.3 (2020-02-29)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 19041)

Matrix products: default

Random number generation:
 RNG:     Mersenne-Twister 
 Normal:  Inversion 
 Sample:  Rounding 
 
locale:
[1] LC_COLLATE=English_United Kingdom.1252  LC_CTYPE=English_United Kingdom.1252    LC_MONETARY=English_United Kingdom.1252 LC_NUMERIC=C                           
[5] LC_TIME=English_United Kingdom.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] openxlsx_4.2.1.9000

loaded via a namespace (and not attached):
[1] compiler_3.6.3 tools_3.6.3    Rcpp_1.0.5     stringi_1.4.6  zip_2.1.1 
```

I also could replicate the problem under Debian
```R
R version 3.6.3 (2020-02-29)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Debian GNU/Linux 9 (stretch)

Matrix products: default
BLAS:   /usr/lib/libblas/libblas.so.3.7.0
LAPACK: /usr/lib/lapack/liblapack.so.3.7.0

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8     LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8    LC_PAPER=en_US.UTF-8      
 [8] LC_NAME=C                  LC_ADDRESS=C               LC_TELEPHONE=C             LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] openxlsx_4.2.1.9000

loaded via a namespace (and not attached):
[1] compiler_3.6.3 tools_3.6.3    Rcpp_1.0.5     stringi_1.5.3  zip_2.1.1     
```

JoshuaSturm:
When grouping a *subset* of columns after setting custom widths,
the index for the `hidden` attribute would start from 1, resulting
in fewer than expected `hidden` values.
It should now correctly match the columns that intersect with those
set by the earlier call to `setColWidths()`.
Fixes #100

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,18 +1,22 @@
 # openxlsx (development version)
 
+## Bug Fixes
+
+* Grouping columns after setting widths no longer throws an error ([#100](https://github.com/ycphs/openxlsx/issues/100))
+
 # openxlsx 4.2.2
 
 ## New Features
 
 * Added features for `conditionalFormatting` to support also 'contains not', 'begins with' and 'ends with'
 
-* Added return value for `saveWorkbook()` the default value for `returnValue` is `FALSE` ([#71])(https://github.com/ycphs/openxlsx/issues/71)
+* Added return value for `saveWorkbook()` the default value for `returnValue` is `FALSE` ([#71](https://github.com/ycphs/openxlsx/issues/71))
 
 * Added Tests for new parameter of `saveWorkbook()`
 
 ## Bug Fixes 
  
-* Solved CRAN check errors based on the change disussed in  (PR#17277)[https://bugs.r-project.org/bugzilla3/show_bug.cgi?id=17277]
+* Solved CRAN check errors based on the change disussed in [PR#17277](https://bugs.r-project.org/bugzilla3/show_bug.cgi?id=17277)
 
 # openxlsx 4.2.0
 
--- a/R/wrappers.R
+++ b/R/wrappers.R
@@ -4292,9 +4292,8 @@ groupColumns <- function(wb, sheet, cols, hidden = FALSE) {
   }
   
   levels <- rep("1", length(cols))
-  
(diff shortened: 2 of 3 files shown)


ycphs:
Thank you

PR #116, opened 2020-11-09
Add arguments to position summary columns and rows

### LINKED ISSUE #107: Setting sheet grouping options (summary rows below/over details; summary columns right/left of details)
By default, Excel's settings for groupings are set to "Summary Rows Below Detail" and "Summary Columns to Right of Detail" but often this is not what is wanted.

**It would therefore be useful to be able to set the grouping options of each sheet.**

* For instance, this could be a function `sheetOptions(wb, sheet, group_summary)` or `sheetGrouping(wb, sheet, summary)` with an argument `group_summary` or `summary` = "BelowRight" / "AboveRight" / "BelowLeft" / "AboveLeft".

This `summary` argument could also be added into the functions `groupColumns` and `groupRows`.

* Note: in order to have the same format of acceptable values as for the `border` argument of the `createStyle` function, a value such as c("below", "right") could also be made acceptable instead of "BelowRight", but I think that this is not very important.

JoshuaSturm:
- You can set the position of summary columns (left/right) and rows (above/below) using the `summaryCol` and `summaryRow` 
arguments, respectively, in `pageSetup()`.

Closes #107 

It's working in Ubuntu + LibreOffice, but corrupting the workbook in Windows + Excel. PR is in draft mode while I debug the Excel issue.


FrancoisR95:
Great! Many thanks @JoshuaSturm for this addition.

ycphs:
I would like to add this feature to the new release to fix the issue #132.

Do you think that's possible?

JoshuaSturm:
@ycphs I hope to finish this weekend.

PR #118, opened 2020-11-18
UTF-8 in/out

### LINKED ISSUE #103: Encoding issue
**Describe the bug**
Encoding issue in sheet names and namedRange

**To Reproduce**
```
library(openxlsx)
Sys.setlocale("LC_ALL", "English")
wb <- createWorkbook()
addWorksheet(wb, "Sheet 1")
addWorksheet(wb, "Šhēet 1")
writeData(wb, sheet = "Sheet 1", x = iris, startCol = 1, startRow = 1)
writeData(wb, sheet = "Šhēet 1", x = iris, startCol = 1, startRow = 1)
createNamedRegion(wb = wb, sheet = "Sheet 1", name = "iris", rows = 1:(nrow(iris) + 1), cols = 1:ncol(iris))

# error
createNamedRegion(wb = wb, sheet = 'Šhēet 1', name = "īriš", rows = 1:(nrow(iris) + 1), cols = 1:ncol(iris))

createNamedRegion(wb = wb, sheet = 'Šhēet 1', name = "iris2", rows = 1:(nrow(iris) + 1), cols = 1:ncol(iris))
out_file <- tempfile(fileext = ".xlsx")
saveWorkbook(wb, out_file, overwrite = TRUE)
getNamedRegions(out_file)
head(read.xlsx(xlsxFile = out_file, namedRegion = "iris"))

# error
head(read.xlsx(xlsxFile = out_file, namedRegion = "iris2"))
```

> createNamedRegion(wb = wb, sheet = 'Šhēet 1', name = "īriš", rows = 1:(nrow(iris) + 1), cols = 1:ncol(iris))
> Error in createNamedRegion(wb = wb, sheet = "Šheet 1", name = "iriš",  : 
>   Invalid characters in name

> head(read.xlsx(xlsxFile = out_file, namedRegion = "iris2"))
> Error in gsub(sheet, "", region, fixed = TRUE) : 
>   invalid 'pattern' argument

**Additional context**
> R version 4.0.2 (2020-06-22)
> Platform: x86_64-w64-mingw32/x64 (64-bit)
> Running under: Windows 10 x64 (build 18362)
> 
> Matrix products: default
> 
> locale:
> [1] LC_COLLATE=English_United States.1252  LC_CTYPE=English_United States.1252   
> [3] LC_MONETARY=English_United States.1252 LC_NUMERIC=C                          
> [5] LC_TIME=English_United States.1252    
> system code page: 1257
> 
> attached base packages:
> [1] stats     graphics  grDevices utils     datasets  methods   base     
> 
> other attached packages:
> [1] openxlsx_4.2.2
> 
> loaded via a namespace (and not attached):
> [1] compiler_4.0.2 tools_4.0.2    Rcpp_1.0.5     stringi_1.5.3  zip_2.1.1  

https://github.com/awalker89/openxlsx/issues/433

shrektan:
Fixes #103

Hi, first of all, thanks for writing this fantasic package.

As a Chinese user, I often need to read / write Excel sheets contain Chinese letters on Windows, where the encoding issue is a huge headache. According to my test, there're at least two known issues::

1. When the comment contains non-ASCII letters, the comment will be garbaged
1. When the Excel table's name contains non-ASCII letters, the name will be garbaged

These two issues are difficult to avoid because by default, Excel will create locale language names for the comments (the author name) and Excel table's names.

And these two issues cause more trouble when I use such an Excel as a template. Resaving the excel object modified by openxlsx will malform the Excel files. I mean, when I open the Excel files that write by openxlsx, Excel will alarm that some contents are illegal.

The root cause of this is that R can't use UTF-8 as the native encoding on Windows as the time of writing (luckily, this may be available in the near future). So, the only reliable cure for this headache for now is to use UTF-8 encoded strings whenever we can. 

This PR tries to do the two things:

1. Ensure all the values of C++ functions that return a character vector be marked as UTF-8 encoded.
1. Defined a helper function `readUTF8()` to ensure that we read the xml as UTF-8 encoded.

Despite it looks like lots of lines has been changed, I believe the PR is simple, if you read the commit in sequence. 

Thanks!

## TODO

- [x] more manual tests on write/read: non-ASCII range names, sheet names, table names, formula
- [x] add unit tests
- [x] add NEWS

---

## Think & Future

We should check all the C++ function's that used in R code and change the argument type from `std::string` to `SEXP`. Only so can we convert the strings to UTF-8 encoded later on the Cpp side, thus **guarantee our policy UTF-8 in / out in everywhere**. The reason is that we can only know the encoding info in `SEXP` but not `std::string`. Another possible solution is to see if there's some configs in Rcpp so that it can ensure the auto-SEXP-to-std::string ends up with a UTF-8 encoded string.

Anyway, I think we'd better not make this PR further complicate and leave this task to the future PRs.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -69,6 +69,7 @@ Collate:
     'openxlsxCoerce.R'
     'readWorkbook.R'
     'sheet_data_class.R'
+    'utils.R'
     'workbook_column_widths.R'
     'workbook_read_workbook.R'
     'workbook_write_data.R'
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -69,7 +69,6 @@ Collate:
     'openxlsxCoerce.R'
     'readWorkbook.R'
     'sheet_data_class.R'
-    'utils.R'
     'workbook_column_widths.R'
     'workbook_read_workbook.R'
     'workbook_write_data.R'
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # openxlsx 4.2.3
  
+## New Features
+
+* Most of functions in openxlsx now support non-ASCII arguments better. More specifically, we can use non-ASCII strings as names or contents for `createNamedRegion()` ([#103](https://github.com/ycphs/openxlsx/issues/103)), `writeComment()`, `writeData()`, `writeDataTable()` and `writeFormula()`. In addition, openxlsx now reads comments and region names that contain non-ASCII strings correctly on Windows. Thanks to @shrektan for the PR [#118](https://github.com/ycphs/openxlsx/pull/118).
+ 
 ## Fixes for Check issues
  
 * Fix to pass the tests for link-time optimization type mismatches
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -37,6 +37,10 @@ get_letters <- function() {
     .Call(`_openxlsx_get_letters`)
 }
 
+markUTF8 <- function(x) {
+    .Call(`_openxlsx_markUTF8`, x)
+}
+
 loadworksheets <- function(wb, styleObjects, xmlFiles, is_chart_sheet) {
     .Call(`_openxlsx_loadworksheets`, wb, styleObjects, xmlFiles, is_chart_sheet)
 }
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -37,8 +37,8 @@ get_letters <- function() {
     .Call(`_openxlsx_get_letters`)
 }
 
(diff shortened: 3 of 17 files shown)


ycphs (changes requested):
I think the function would also fit into the helper.R file

ycphs (commented):
I am currently planning to replace all the `readLines` with an equivalent in cpp. That should boost the performance and the UTF8 encoding can be solved at the same time.

ycphs:
Thank you for your work. This will improve the package a lot.

PR #121, opened 2020-11-28
Implement reading namedRegion from specific sheet

JanMarvin:
Frist: I'm sorry for jumping the line and starting with a pull request right away. Though I needed the following issue solved and therefore had to write the code either way.

## The issue:

The xlsx format allows having identically named regions on multiple sheets. Previously `openxlsx` would ignore such cases and simply jump to the first appearance of a `namedRegion` even if a sheet other than `1` is selected. The output does not inform the user of such a case but the following if condition will trow a warning
https://github.com/ycphs/openxlsx/blob/610d843fdb5dc8bbff143b5068a3edd397aac7dc/R/readWorkbook.R#L252

My initial idea was a single line edit to silence the warning. Tough the real issue is that only the first appearance of a `namedRegion` can be returned. 

## This PR solve the issue at hand as follows:

* obtain all the sheet names with a selected `namedRegion` in `dn_sheetNames` and compare this with the `sheetNames` to get the ids of sheets with this `namedRegion`
* if multiple sheets contain the namedRegion print a message to inform the user
* if the user has selected a `namedRegion` the selected sheet will be used as data source

Since `openxlsx` defaults on `sheet = 1` the following corner cases may occur producing unpleasant output

* if the user has selected a `namedRegion`, but a wrong sheet, a message will inform the user and the first appearance will be selected (for backward comparability)
* if the user has selected `sheet == 1` a message will appear

This could be solved if `openxlsx` would not default to 1 and instead would use something like `if (missing(sheet)) sheet == 1` if no sheet was selected.

I added a test file to the pull request. This xlsx was created with Libre Office 7.0.3.1, but similar things can be done using MS Office. Maybe even with `openxlsx`, but I'm not familiar enough with the code base.


ycphs:
Thank you for your contribution!

JanMarvin:
This pull request solved an issue, but isn't general enough that every xlsx file with `namedRegion`s will work. I'm currently trying to solve this issue and once finished will create a follow up pull request.  Tackling this is a bit annoying, as some issue arise only with specific xlsx files I am unable to share and can not really use for testing nor reproduce locally. For starters: you can have a `namedRegion` on a linked excel sheet, yay!

PR #123, opened 2020-12-04
Further improvements regarding named regions

JanMarvin:
As stated in #121 the initial fix is an improvement to the not really functional previous situation, though it failed with some xlsx files I am currently dealing with. Said files have numerous named regions. These include references to external xlsx files and contain already removed and/or hidden sheets. These files have been worked with heavily over the last few years. With such files, my initial implementation failed.
And since I wasn't familiar enough with how `openxlsx` works, I added a bug myself. The current implementation only works with the sheet id, not with a character string.

This is solved in this follow up pull request. I have marked it work in progress since it needs to be rebased and cleaned up (e.g., remove debug and unintended .github merge). I have begun testing it over the last few days. For now I would like to ask for feedback:

* this pull request is no longer entirely backward compatible
   - it sets the default for `sheet` as missing: previously it was impossible to identify if a sheet was specifically selected. Maybe the sheet id was 1 as selected or 1 by default.
   - its behavior is no longer entirely backward compatible: it does no longer return something, if the namedRegion is valid, but the sheet is wrong. Since I loop over some sheets, I would end up with a variety of results if the namedRegion was found, even though not on the selected sheet.
* if these changes are unwanted, I would still push a fix for the handling of character `sheet`.

* lastly I would like to ask for feedback. maybe someone has a file to test or at least runs a mac/windows with office (to avoid the introduction of new bugs :smile: )


ycphs:
Thank you once more for the contribution. I know about the importance of backward compatibility, but in this case it seems better to have this new features in. 

For me it is fine, because I don't need named regions that often. Maybe it could be useful to deprecate `namedRegion` and use a slightly different like `NamedRegion` or just `region`. This would then allow a smooth transition and give other developers the chance to adapt their code.

What are your thoughts @JanMarvin ?

JanMarvin:
Since this particular feature has not been functioning properly previously, I generally assume the impact is negligible. Especially giving the improvement of having a functioning `namedRegion`. Though I wanted to be crystal clear regarding the changes since the onus of dealing with user complaints will be on you 😉 

Implementing `sheet` as `new_sheet` or something alike as you suggested might be feasible (`namedRegion` itself is not actually a problem). Though I think breaking backward compatibility regarding this issue is favourable.

Since `openxlsx` offers interfaces for files and for workbooks: my PR has and will be tested with files exclusively since I do not need workbooks. As far as I can tell, `namedRegion` functionality for workbooks is not on par with files, but improvements for workbooks will not be a part of this pull request.

To conclude: I will clean up and rebase this PR and look if I can come up with a backward compatible solution. I hope to have finished this and a bit more testing until next weekend.

JanMarvin:
Sadly I have to postpone the rebase and testing for a week. It's year-end crunch time and I could not do a thing over the week.

ycphs:
No problem, there is no rush. I always try to merge the PRs as soon as possible to the master.

JanMarvin:
I have now cleaned the pull request. Meaning I have removed the debug output and added more comments to the code. 

* Even though all tests complete (as they did before), it was not tested as much as I wanted to (had other things on my table, but likely will fetch it up early next year).
* I have decided *against* keeping backward comparability (not just because I was lazy :wink: ), but because I feel that the user should not expect correct output with incorrect input. If `openxlsx` is used in a process that relies on this old behavior, it is a) likely that is on an otherwise unmaintained system, where it does not matter at all (aka it will not be updated anyway), or b) it should be fairly easy to adapt to the new behavior and c) it might show previously incorrect results. I got the impression, that you agree with me regarding this matter, but if required, I'd be willing to switch that.

[Edit:] I can see that a codecov check is failing. Though, I cannot see why or what is going on. Please let me know if there is anything I can do to fix this.

Issue #124, opened 2020-12-10
Error loading workbook with pivot tables


JMPivette:
**Describe the bug**
When loading a file where the number of pivot tables is not a multiple of the pivotCacheDefinition's number we get an error.
This error usually happens when multiple pivot tables refer to the same source.

**To Reproduce**
1. [template_error.xlsx](https://github.com/ycphs/openxlsx/files/5675780/Book3.xlsx)
2. Loading _template_error.xlsx_ with loadWorkbook() throws an error:

```
Error in sprintf("<Relationship Id=\"rId%s\" Type=\"http://schemas.openxmlformats.org/officeDocument/2006/relationships/pivotCacheDefinition\" Target=\"pivotCache/pivotCacheDefinition%s.xml\"/>",  : 
  arguments cannot be recycled to the same length
```
This template file has 3 pivot tables but 2 of them are referring to the same data.

**Desktop (please complete the following information):**
 - Version [4.2.3]

**Additional context**

Bug is also described here:
https://github.com/awalker89/openxlsx/issues/387
https://github.com/awalker89/openxlsx/issues/377

JMPivette:
A solution had been created here in 2018: https://github.com/awalker89/openxlsx/pull/417/commits/63bec3ac735e3c26d2934e7870eaa3676353cfc2

But I am not sure it is the best way to solve it

PR #125, opened 2020-12-11
Fix issue #124

JMPivette:
I think I found the problem for issue #124 . 
The code was referring to the number of `pivotTables` where it was supposed to refer to the number of `pivotCacheDefinition`.

Most of the time these numbers are equivalent so there was no errors. 
Also when the number of `pivotTables` were a multiple of `pivotCacheDefinition`, the values were recycled by `sprintf` without any warning.

I also added a file and a unit test corresponding to issue #124 

R CMD check succeeded and I did additional manual testing to see if I could still load and save different types of xlsx files.
However I am not an Excel expert, so a review is welcome 😃


tests/testthat/test-loading_workbook.R
```diff
@@ -902,3 +902,10 @@ test_that("Loading readTest.xlsx Sheet 1", {
 
   expect_equal(expected_shared_strings, wb$sharedStrings)
 })
+
+test_that("Loading multiple pivot tables: loadPivotTables.xlsx works",{
+  ## loadPivotTables.xlsx is a file with 3 pivot tables and 2 of them have the same reference data.
+  fl <- system.file("extdata", "loadPivotTables.xlsx", package = "openxlsx")
+  wb <- loadWorkbook(fl)
+
```
JanMarvin, line 910: FYI the actual test is missing. This should most likely include any kind of `test_` for `wb` e.g., is the data correctly loaded. I'm not even sure if testthat does complain if loadworkbook failed.

ycphs: @JanMarvin yes you are correct. It needs something like `expect_equal()`.

ycphs:
Thank you for your PR

tests/testthat/test-loading_workbook.R
JMPivette: @JanMarvin yes you are right, it will be cleaner to have an `expect_` function in the test. I will add it.

For information, if any error occurs inside test_that() function, testthat raises an error.
This test without the code modification creates an error.

PR #130, opened 2020-12-21
[WIP] improve getChildlessNode. fixes #127

### LINKED ISSUE #127: saveWorbook() creates "corrupted" excel file when autoFilter is used
Same bug as:
https://github.com/awalker89/openxlsx/issues/524

**Describe the bug**
If autoFilter is use inside an Excel file, saveWorkbook() will create a "corrupted" file.

**To Reproduce**
1. Use `loadWorkbook()` on this file: 
[auto_filter_errors.xlsx](https://github.com/ycphs/openxlsx/files/5688356/auto_filter_errors.xlsx)

2. Save the workbook using `saveWorkbook()`
3. Open the created file with Excel
4. Excel gives the following error:
`We found a problem with some content in ’filter_output.xlsx’. Do you want us to try to recover as much as we can? If you trust the source of this workbook, click Yes.`
Looking at the logs:
`<removedParts summary="Following is a list of removed parts:"><removedPart>Replaced Part: /xl/worksheets/sheet1.xml part with XML error.  Xml parsing error Line 1, column 1518.</removedPart>`

**Desktop :**
 - R: [4.0.2]
 - Version [4.2.3]

JanMarvin:
This pull request improves getChildlessNode. Previously it would only work flawlessly with `<foo bar/>` and not with `<foo><bar/></foo>`. The latter would read only `<foo><bar/>` resulting in incorrect xlsx files when written from a workbook.

```R
xml <- "<html is not a programming language><a ><b /></a><b /><a ></a><c /><yes it is a programming language>"
```

With this pull request it will be:

```R
> openxlsx:::getChildlessNode(xml, "a")
[1] "<a ><b /></a>" "<a ></a>"     
> openxlsx:::getChildlessNode(xml, "b")
[1] "<b />" "<b />"
> openxlsx:::getChildlessNode(xml, "c")
[1] "<c />"
``` 

Currently in master it is:

```R
> openxlsx:::getChildlessNode(xml, "<a")
[1] "<a ><b />"     "<a ></a><c />"
> openxlsx:::getChildlessNode(xml, "<b")
[1] "<b />" "<b />"
> openxlsx:::getChildlessNode(xml, "<c")
[1] "<c />"
```

This should be testet. In theory it should be slower than the current implementation, due to the initial check. It requires a few unit tests and the written xlsx files should be tested with Microsofts Excel since the issue does not arise with LibreOffice or R itself. I was able to run the unit tests prior to the completion of the syntax switch.

### DIFF
--- a/R/WorkbookClass.R
+++ b/R/WorkbookClass.R
@@ -17859,7 +17859,7 @@ Workbook$methods(
 
 
     ## Number formats
-    numFmts <- getChildlessNode(xml = stylesTxt, tag = "<numFmt ")
+    numFmts <- getChildlessNode(xml = stylesTxt, tag = "numFmt")
     numFmtFlag <- FALSE
     if (length(numFmts) > 0) {
       numFmtsIds <-
@@ -17907,7 +17907,7 @@ Workbook$methods(
 
     cellXfs <- getNodes(xml = stylesTxt, tagIn = "<cellXfs")
 
-    xf <- getChildlessNode(xml = cellXfs, tag = "<xf ")
+    xf <- getChildlessNode(xml = cellXfs, tag = "xf")
     xfAttrs <- regmatches(xf, gregexpr('[a-zA-Z]+=".*?"', xf))
     xfNames <-
       lapply(xfAttrs, function(xfAttrs) {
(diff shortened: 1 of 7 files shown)


JanMarvin:
I agree with @JMPivette about the different logic of getChildlessNode and getNode. My approach was to create a unification that works independent of the input. This might be worth looking into, if the separation is wanted (might be beneficial regarding runtime) or if we should get rid of it entirely.

PR #131, opened 2020-12-21
[WIP] Writing Array Functions

### LINKED ISSUE #50: Excel Array Formulas
**Is your feature request related to a problem? Please describe.**
I'm trying to automate the creation of a survey workbook that incorporates Excel Tables for data input. I currently manually create array formulas to add an automatically-updated visible entry number to the table. My workbook now has dozens of tables, so the array formula creation is becoming burdensome.

**Describe the solution you'd like**
I would like to be able to use openxlsx to create these array formulas as part of my automation.

**Describe alternatives you've considered**
I cannot think of alternatives other than continuing manual array formula creation.

JanMarvin:
This is a working implementation to further increase `writeFormula()` to be able to write array functions. This closes #50

Right now the code is a bit ugly, since I was mostly duplicating existing code, other than that it works as expected.

Using the example code from #114 (which I could not confirm) the following is now possible:

```R
wb <- createWorkbook()
addWorksheet(wb,'Test')
testMat <- matrix(rep(10,10),10,1)
writeData(wb,'Test',testMat)

testFormula <- 'SUM(ABS(A2:A11))'
writeFormula(wb,'Test',testFormula,startCol = 1,startRow = 13, array = F)  # old default, does not work without array
writeFormula(wb,'Test',testFormula,startCol = 1,startRow = 14, array = T)  # new
writeFormula(wb,'Test',testFormula,startCol = 4,startRow = 4, array = T)   # new just in a different cell

saveWorkbook(wb,'test2.xlsx',overwrite = T)
```

Though it is ~~currently~~ not possible to import the written formulas back into R.

ToDo:

* [x] clean up (remove debug, naming vector_ -> array_)
* [x] testing with Excel

### DIFF
--- a/R/workbook_write_data.R
+++ b/R/workbook_write_data.R
@@ -84,17 +84,19 @@ Workbook$methods(writeData = function(df, sheet, startRow, startCol, colNames, c
     }
   }
 
-  if ("formula" %in% allColClasses) {
-    for (i in which(sapply(colClasses, function(x) "formula" %in% x))) {
-      df[[i]] <- replaceIllegalCharacters(as.character(df[[i]]))
-      class(df[[i]]) <- "openxlsx_formula"
+  if (any(c("formula", "array_formula") %in% allColClasses)) {
+    
+    frm <- "formula"
+    cls <- "openxlsx_formula"
+    
+    if ("array_formula" %in% allColClasses) {
+      frm <- "array_formula"
+      cls <- "openxlsx_array_formula"
     }
-  }
-
-  if ("array_formula" %in% allColClasses) {
-    for (i in which(sapply(colClasses, function(x) "array_formula" %in% x))) {
+    
+    for (i in which(sapply(colClasses, function(x) frm %in% x))) {
       df[[i]] <- replaceIllegalCharacters(as.character(df[[i]]))
-      class(df[[i]]) <- "openxlsx_array_formula"
+      class(df[[i]]) <- cls
     }
   }
 
--- a/R/workbook_write_data.R
+++ b/R/workbook_write_data.R
@@ -175,7 +175,7 @@ Workbook$methods(writeData = function(df, sheet, startRow, startCol, colNames, c
   any_functions <- FALSE
   ref_cell <- paste0(int_2_cell_ref(startCol), startRow)
 
-  if (c("openxlsx_formula", "openxlsx_array_formula") %in% colClasses) {
+  if (any(c("openxlsx_formula", "openxlsx_array_formula") %in% colClasses)) {
     
     ## alter the elements of t where we have a formula to be "str"
     if ("openxlsx_formula" %in% colClasses) {
--- a/R/workbook_write_data.R
+++ b/R/workbook_write_data.R
@@ -85,9 +85,18 @@ Workbook$methods(writeData = function(df, sheet, startRow, startCol, colNames, c
   }
 
   if ("formula" %in% allColClasses) {
+    print("Formula")
     for (i in which(sapply(colClasses, function(x) "formula" %in% x))) {
       df[[i]] <- replaceIllegalCharacters(as.character(df[[i]]))
       class(df[[i]]) <- "openxlsx_formula"
+    }
+  }
+
+  if ("vector_formula" %in% allColClasses) {
+    print("Vector Formula")
+    for (i in which(sapply(colClasses, function(x) "vector_formula" %in% x))) {
+      df[[i]] <- replaceIllegalCharacters(as.character(df[[i]]))
+      class(df[[i]]) <- "openxlsx_vector_formula"
     }
   }
 
@@ -163,15 +172,31 @@ Workbook$methods(writeData = function(df, sheet, startRow, startCol, colNames, c
   }
 
 
+  ref_cell <- paste0(int_2_cell_ref(startCol), startRow)
 
-  ## Forumlas
+  ## Formulas
   f_in <- rep.int(as.character(NA), length(t))
   any_functions <- FALSE
+
   if ("openxlsx_formula" %in% colClasses) {
 
     ## alter the elements of t where we have a formula to be "str"
     formula_cols <- which(sapply(colClasses, function(x) "openxlsx_formula" %in% x, USE.NAMES = FALSE), useNames = FALSE)
-    formula_strs <- stri_join("<f>", unlist(df[formula_cols], use.names = FALSE), "</f>")
+    formula_strs <- stri_join("<f aca=\"false\">", unlist(df[formula_cols], use.names = FALSE), "</f>")
+    formula_inds <- unlist(lapply(formula_cols, function(i) i + (1:(nRows - colNames) - 1) * nCols + (colNames * nCols)), use.names = FALSE)
+    f_in[formula_inds] <- formula_strs
+    any_functions <- TRUE
+
+    rm(formula_cols)
+    rm(formula_strs)
+    rm(formula_inds)
+  }
+
+  if ("openxlsx_vector_formula" %in% colClasses) {
+
+    ## alter the elements of t where we have a formula to be "str"
+    formula_cols <- which(sapply(colClasses, function(x) "openxlsx_vector_formula" %in% x, USE.NAMES = FALSE), useNames = FALSE)
+    formula_strs <- stri_join("<f aca=\"false\" t=\"array\" ref=\"", ref_cell, ":", ref_cell, "\">", unlist(df[formula_cols], use.names = FALSE), "</f>")
     formula_inds <- unlist(lapply(formula_cols, function(i) i + (1:(nRows - colNames) - 1) * nCols + (colNames * nCols)), use.names = FALSE)
     f_in[formula_inds] <- formula_strs
     any_functions <- TRUE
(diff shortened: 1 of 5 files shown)


R/workbook_write_data.R
```diff
@@ -163,15 +172,31 @@ Workbook$methods(writeData = function(df, sheet, startRow, startCol, colNames, c
...
+  ## Formulas
   f_in <- rep.int(as.character(NA), length(t))
   any_functions <- FALSE
+
   if ("openxlsx_formula" %in% colClasses) {
 
     ## alter the elements of t where we have a formula to be "str"
     formula_cols <- which(sapply(colClasses, function(x) "openxlsx_formula" %in% x, USE.NAMES = FALSE), useNames = FALSE)
-    formula_strs <- stri_join("<f>", unlist(df[formula_cols], use.names = FALSE), "</f>")
+    formula_strs <- stri_join("<f aca=\"false\">", unlist(df[formula_cols], use.names = FALSE), "</f>")
```
JanMarvin: no clue what this aca="false" statement does. have seen it with libreoffice files. presumably it can be dropped again

R/workbook_write_data.R
```diff
@@ -163,15 +172,31 @@ Workbook$methods(writeData = function(df, sheet, startRow, startCol, colNames, c
...
+    formula_inds <- unlist(lapply(formula_cols, function(i) i + (1:(nRows - colNames) - 1) * nCols + (colNames * nCols)), use.names = FALSE)
+    f_in[formula_inds] <- formula_strs
+    any_functions <- TRUE
+
+    rm(formula_cols)
+    rm(formula_strs)
+    rm(formula_inds)
+  }
+
+  if ("openxlsx_vector_formula" %in% colClasses) {
```
JanMarvin: I'll see if this can be integrated into the "normal one above"

R/workbook_write_data.R
```diff
@@ -85,12 +85,21 @@ Workbook$methods(writeData = function(df, sheet, startRow, startCol, colNames, c
...
 
   if ("formula" %in% allColClasses) {
+    print("Formula")
     for (i in which(sapply(colClasses, function(x) "formula" %in% x))) {
       df[[i]] <- replaceIllegalCharacters(as.character(df[[i]]))
       class(df[[i]]) <- "openxlsx_formula"
     }
   }
 
+  if ("vector_formula" %in% allColClasses) {
```
JanMarvin: should be array_formula

ycphs:
I would like to add this feature to the new release to fix the issue #132.

Do you think that's possible?

JanMarvin:
> I would like to add this feature to the new release to fix the issue #132.
> 
> Do you think that's possible?

I'm not sure what you mean by this solving #132. This adds an entirely new feature - most likely unrelated to the failing check. Could you please clarify what you mean?

On the pr side of things, I believe that this pr, with a bit of polishing, could be ready for the next release. Just let me know when you intend to release and I'll make some time for it.

ycphs:
> 
> I would like to add this feature to the new release to fix the issue #132.
> 
> Do you think that's possible?

Add the feature and fix the issue #132. 

Sorry for the mid understanding.

R/writeData.R
```diff
@@ -226,12 +228,21 @@ writeData <- function(wb,
     colNames <- FALSE
   }
 
+if (!array) {
   ## special case - formula
   if ("formula" %in% class(x)) {
     x <- data.frame("X" = x, stringsAsFactors = FALSE)
     class(x[[1]]) <- "formula"
```
JanMarvin: ifelse

R/writeData.R
```diff
@@ -499,13 +511,19 @@ writeFormula <- function(wb,
...
                          xy = NULL) {
   if (!"character" %in% class(x)) {
     stop("x must be a character vector.")
   }
 
-  dfx <- data.frame("X" = x, stringsAsFactors = FALSE)
-  class(dfx$X) <- c("character", "formula")
+  if (!array) {
+    dfx <- data.frame("X" = x, stringsAsFactors = FALSE)
+    class(dfx$X) <- c("character", "formula")
```
JanMarvin: ifelse

JanMarvin:
`# nolint start`

JanMarvin:
Needs a check with Excel (will test it next week) and possibly a unit test. Though, since we can not read formulas, a missing will be imported. The only possible workaround would be a test, if the written file contains the expected attributes. I'm not entirely keen on writing such test, but currently I lack the imagination, how else this could be tested.

Otherwise it is good to go.

JanMarvin:
Tested with the Android Excel. Please let me know, if you want the test.

inst/extdata/build_font_size_lookup.R
```diff
@@ -1,3 +1,4 @@
+# nolint start
```
JanMarvin, line 1: *sigh* this should be added to all extdata R files

ycphs:
great work thank you

khorovatin:
Thank you for adding this feature! I look forward to trying it out.

khorovatin:
Apologies for taking so long to actually trying the feature--I only generate the spreadsheets that use this package a couple times a year, and only just started the next run.
Sadly, for my use case, this doesn't seem to quite make it: I still have to open the file in Excel, select the cell with the array formula, and press `<Shift>+<Ctrl>+<Enter>` to have the formula behave the way I expect.
I installed the development version (4.2.3.9000) this morning, using R 4.0.5.
I've attached a short script file to illustrate the problem. I create a table, then paste the formula in the first column of the first data row. The formula should create an auto-incrementing "id" each time the user adds a new row to the table.
What happens is that the first row correctly gets the id set to "1", but new rows (added by hitting tab while the "name" column of the existing row is the active cell) do not inherit the formula. I need to highlight the first row's "id" cell, shift focus to the formula bar, and hit the `<Shift>+<Ctrl>+<Enter>` key combination to create the correct array formula, which then auto-populates to the new rows below and correctly calculates an incrementing "id" value.
[array_formula.txt](https://github.com/ycphs/openxlsx/files/6260535/array_formula.txt)

JanMarvin:
Hi Ken, the following works for me. I've simply removed this part: ` class(id_formula) <-   c(class(id_formula), "formula")`

LibreOffice evaluates it as a valid array formula:

Screenshot_2021-04-05_23-12-59

Your code looks like this. Please let us know if this works for you too.

```r
library(tibble)
library(openxlsx)
library(glue)

wb <- createWorkbook(
  creator = "khorovatin",
  title = "Test Array Formula"
)

sheet_name <- "Table With Array"
sheet_no <- 1
start_row <- 2
input_table <- tribble(
  ~id, ~name,
)

addWorksheet(wb, sheet_name)

# Input Table
writeDataTable(
  wb,
  sheet_no,
  input_table,
  startCol = 2,
  startRow = start_row,
  tableStyle = "TableStyleMedium15",
  tableName = "input_table",
  withFilter = FALSE
)

# Set up id column formula
# glue is not required: id_formula <- "ROW()-ROW(input_table[[#Headers],[Id]])"
id_formula <-
  as.character(
    glue(
      '
        ROW()-ROW(input_table[[#Headers],[Id]])
            '
    )
  )

writeFormula(wb,
             sheet_no,
             id_formula,
             startCol = 2,
             startRow = start_row + 1,
             array = TRUE)

openXL(wb)
```

khorovatin:
[file32246a25f2b.xlsx](https://github.com/ycphs/openxlsx/files/6261562/file32246a25f2b.xlsx)
Thanks for your reply. I removed the `class(id_formula) <- c(class(id_formula), "formula")` line, but saw no change in behavior.
I'm still not getting the array formula copying to new rows when they're created. (Note, I am using the Microsoft 365 version of Excel):
image
Two new rows:
image
And after highlighting the first "id" row and pressing `<Shift>+<Ctrl>+<Enter>`:
image

Finally thanks for removing the `glue`. I missed removing that--my original code has embedded variables in the formula and I forgot to remove it from my sample code.

Edit: I just opened the file in LibreOffice and see that copying the rows down _does_ use the array formula the way I expect, but LibreOffice does not seem to have support for Excel Tables. I wonder if the Excel Table has something to do with the problem?
I've attached the `xlsx` file containing the formula _after_ pressing `<Shift>+<Ctrl>+<Enter>`, in case it helps.

JanMarvin:
thanks for the follow up and the attached file, I'll look into it, but it might take a few days

JanMarvin:
Thanks again for the file. Guess you're correct pointing at the table. I assume we'd have to include something like this

```xml
<calculatedColumnFormula array="1">ROW()-ROW(input_table[[#Headers],[id]])</calculatedColumnFormula>
```

in node `tableColumn` 

of https://github.com/ycphs/openxlsx/blob/686347602af1506b2e7adad56c5241825b485dae/src/write_file.cpp#L326

used in 

https://github.com/ycphs/openxlsx/blob/686347602af1506b2e7adad56c5241825b485dae/R/WorkbookClass.R#L1171-L1173

when calling

https://github.com/ycphs/openxlsx/blob/686347602af1506b2e7adad56c5241825b485dae/R/writeDataTable.R#L138-L140

At least that's what it looks like from your file.

I suggest adding `calculateColumnFormula` as character vector to `writeDataTable` and a similar bool like vector (TRUE, FALSE, NA)  indicating (array) formulas. Once these are passed down to the cpp function, a simple check for the bool should indicate if we set `array = "1"` or add the calculatedColumnFormula node.  Since I do not own Excel and can not test this, maybe someone else could step up adding these? The patch should be rather straight forward (I assume that I've basically added everything required), though I'm not entirely sure that that's sufficient.

Issue #112, opened 2020-10-21
read.xlsx skips strings when reading file generated by openpyxl (Python)


bersbersbers:
**Describe the bug**
The attached file `broken4.xlsx` looks like this in Excel:
![image](https://user-images.githubusercontent.com/12128514/96745787-cc7b0a00-13c6-11eb-9741-2797893070ee.png)
[broken4.xlsx](https://github.com/ycphs/openxlsx/files/5417241/broken4.xlsx)

However, when read using `openxlsx::read.xlsx("broken4.xlsx")`, I only get this:
```
> openxlsx::read.xlsx("broken4.xlsx")
  1 10000 10000 -1.796071094480823
1 1 10000  3000          1.0102900
2 1 10000  1000          0.4490178
3 1 10000   300          0.0000000
4 1 10000   100          0.8980355
5 1 10000    30          0.1122544
6 1 10000    10          3.1431244
7 1 10000     0          1.4593078
8 2 10000     0          1.6838167
9 2 10000    10         -0.7857811
```

**To Reproduce**
1. `openxlsx::read.xlsx("broken4.xlsx")`

**Expected behavior**
When opening the file in Excel, resaving it, and then opening it, I get this:
```
> openxlsx::read.xlsx("resave.xlsx")
   PairIndex Drug1 Drug2 Conc1 Conc2   Response ConcUnit
1          1   abc   def 10000 10000 -1.7960711       nM
2          1   abc   def 10000  3000  1.0102900       nM
3          1   abc   def 10000  1000  0.4490178       nM
4          1   abc   def 10000   300  0.0000000       nM
5          1   abc   def 10000   100  0.8980355       nM
6          1   abc   def 10000    30  0.1122544       nM
7          1   abc   def 10000    10  3.1431244       nM
8          1   abc   def 10000     0  1.4593078       nM
9          2   abc   def 10000     0  1.6838167       nM
10         2   abc   def 10000    10 -0.7857811       nM
```

**Screenshots**

**Additional context**
Note that `broken4.xlsx` is a combination of the original `broken.xlsx`, in which I, one by one, replaced all files different between `resave.xlsx` and `broken.xls`. The only files different between the two are in `.\xl\worksheets\sheet1.xml`, which look like this:

`broken4.xlsx`:
<details>

```
[...]

PR #134, opened 2021-01-02
[WIP] return file readable for Excel. closes #81

### LINKED ISSUE #77: createStyle with insert backslash in numFmt
Hello,
Thanks a lot for your work.

When I run the code :

```
library(openxlsx)
wb <- createWorkbook(title = "debug.xlsx")
addWorksheet(wb = wb, sheetName = "debug")
writeData(wb = wb, sheet = 1, x = 12456789, startCol = 1, startRow = 1)
style_euros = createStyle(numFmt="# ##0 €")
addStyle(wb = wb, sheet = 1, rows = 1, cols = 1, style = style_euros)
saveWorkbook(wb = wb, file = "debug.xlsx", overwrite = TRUE)
```

Back slash are inserted into my format :
currency_style

Thus, I get `12456 789 €` instead of `12 456 789 €`

My sessionInfo :
```
R version 3.6.3 (2020-02-29)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 18.04.4 LTS

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.7.1
LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.7.1

locale:
 [1] LC_CTYPE=fr_FR.UTF-8       LC_NUMERIC=C               LC_TIME=fr_FR.UTF-8        LC_COLLATE=fr_FR.UTF-8     LC_MONETARY=fr_FR.UTF-8   
 [6] LC_MESSAGES=fr_FR.UTF-8    LC_PAPER=fr_FR.UTF-8       LC_NAME=C                  LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=fr_FR.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] openxlsx_4.1.5.1

loaded via a namespace (and not attached):
[1] compiler_3.6.3 tools_3.6.3    Rcpp_1.0.4.6   stringi_1.4.6  zip_2.0.4 
```

Thank you

### LINKED ISSUE #81: writing to existing excel workbook corrupts excel file
**Describe the bug**
First of all, thanks more making such a great package. I used it last year and it was an amazing help. This is my first bug posting, so please forgive me if I am missing any relevant details. 

I am writing data to an existing excel workbook. In the old version of the workbook, which DID contain some data validations etc, it worked fine. I was able to open the file after running it through openxlsx with no issues. I am now working with a very similar excel workbook this year, nothing major changed in it, but for some reason no matter what I do, even creating another version of the file, I get an error message in excel saying the file is corrupted and unrepairable after running it through openxlsx. Running the old excel file, again no issues at all. The file sizes are about the same between the old and new templates. Again, nothing major got added to this new template that I would have reason to believe is interfering. I have searched through all of the openxlsx help files, and no matter how I change the template cannot find the source of the problems. 

**To Reproduce**
I am working in latest openxlsx development package version 4.1.5.1. 
1. Here are the two workbook template files: 
[Workbook Template new.xlsx](https://github.com/ycphs/openxlsx/files/4879159/Workbook.Template.new.xlsx)
[Workbook Template old.xlsx](https://github.com/ycphs/openxlsx/files/4879163/Workbook.Template.old.xlsx)

test=c("3", "2.5", "1")
wb <- openxlsx::loadWorkbook("Workbook Template new.xlsx") 
#One row of data just to illustrate problem. 
openxlsx::writeData(wb,"Basic inputs", test[1], startRow=2,startCol=3, colNames=F)
openxlsx::saveWorkbook(wb, file=paste("test", "abc",".xlsx",sep=""), overwrite=T)
#I get error messages here saying the file has been corrupted. 

**Expected behavior**
Expected result occurs when I do the following with the old workbook. 
test=c("3", "2.5", "1")
wb <- openxlsx::loadWorkbook("Workbook Template old.xlsx") 
#One row of data just to illustrate problem. 
openxlsx::writeData(wb,"Basic inputs", test[1], startRow=2,startCol=3, colNames=F)
openxlsx::saveWorkbook(wb, file=paste("test", "abc",".xlsx",sep=""), overwrite=T)
#This file opens fine. 

**Screenshots**
Below are the errors I get when trying to open the excel file. 
image
image

**Desktop (please complete the following information):**
R version 3.5.1 (2018-07-02)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows >= 8 x64 (build 9200)
Matrix products: default
Browser: Chrome

locale:
[1] LC_COLLATE=English_Canada.1252 
[2] LC_CTYPE=English_Canada.1252   
[3] LC_MONETARY=English_Canada.1252
[4] LC_NUMERIC=C                   
[5] LC_TIME=English_Canada.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets 
[6] methods   base     

other attached packages:
 [1] openxlsx_4.1.5.1 zip_2.0.4        Rcpp_1.0.4.6    
 [4] naniar_0.5.2     fuzzyjoin_0.1.6  plyr_1.8.4      
 [7] stringr_1.3.1    assertr_2.6      rJava_0.9-10    
[10] RODBC_1.3-15     lattice_0.20-38  tidyr_1.1.0     
[13] devtools_2.3.0   fansi_0.4.1      usethis_1.6.1   
[16] dplyr_0.8.5     

loaded via a namespace (and not attached):
 [1] tidyselect_1.1.0  remotes_2.1.1    
 [3] purrr_0.3.4       colorspace_1.4-1 
 [5] vctrs_0.3.1       testthat_2.3.2   
 [7] yaml_2.2.1        rlang_0.4.6      
 [9] pkgbuild_1.0.8    pillar_1.4.4     
[11] glue_1.4.1        withr_2.2.0      
[13] sessioninfo_1.1.1 lifecycle_0.2.0  
[15] munsell_0.5.0     gtable_0.3.0     
[17] memoise_1.1.0     callr_3.4.3      
[19] ps_1.3.3          curl_4.3         
[21] backports_1.1.8   scales_1.0.0     
[23] desc_1.2.0        pkgload_1.1.0    
[25] fs_1.4.2          ggplot2_3.1.0    
[27] digest_0.6.25     stringi_1.4.6    
[29] visdat_0.5.3      processx_3.4.2   
[31] rprojroot_1.3-2   grid_3.5.1       
[33] cli_2.0.2         tools_3.5.1      
[35] magrittr_1.5      lazyeval_0.2.2   
[37] tibble_3.0.1      crayon_1.3.4     
[39] pkgconfig_2.0.3   ellipsis_0.3.1   
[41] prettyunits_1.1.1 assertthat_0.2.1 
[43] rstudioapi_0.11   R6_2.4.1         
[45] compiler_3.5.1   

**Additional context**
Any suggestions about how I can alter the template to be able to work with openxlsx would be much appreciated. I am at a complete loss, I've tried changing the file, I've updated all R packages, I've ensured RTools40 is working properly with my R, but nothing seems to fix this corruption issue. So this leads me to believe its some sort of bug. Thanks in advance. Happy to provide any further information!

Dorothy

JanMarvin:
Early work in progress. Should fix #81 

* [ ] file loads in excel

My idea is to minimize the differences between the original file and the one returned by `openxlsx`. Thus if Excel works with one it will work with the other as well. Since Excel won't load the file, the status is tracked mostly using libre office which loads the file.

What this PR does so far is modifying the import logic. It should keep the fonts of the original xlsx file intact. This requires a bit of digging around the codebase as there are a few stones in our way. One being the lack of a good XML parser (apparently half the codebase is trying to get content from the various XML sheets). Once the layout is identical between the two files, hopefully the issue is closed.

* [x] import fonts
* [x] keep protect intact
* [x] keep the layout of the sheets
* [ ] other things

This is also an invitation for others to test what is working and what is not. If you want to join in, unzip the xlsx files and compare the XML files. Open up visual studio code and start searching for strings.


R/WorkbookClass.R
```diff
@@ -17905,9 +17905,9 @@ Workbook$methods(
 
     ## ------------------------------ build styleObjects ------------------------------ ##
 
-    cellXfs <- getNodes(xml = stylesTxt, tagIn = "<cellXfs")
-
-    xf <- getChildlessNode(xml = cellXfs, tag = "xf")
+    cellXfs <<- getChildlessNode(xml = stylesTxt, tag = "cellXfs")
+    xf <<- getNodes(xml = cellXfs, tagIn = "<xf")
```
JanMarvin: I might fix getNodes or better, I might try to get what I want from it

R/loadWorkbook.R
```diff
@@ -365,6 +365,13 @@ loadWorkbook <- function(file, xlsxFile = NULL, isUnzipped = FALSE) {
   ## xl\styles
   if (length(stylesXML) > 0) {
     styleObjects <- wb$loadStyles(stylesXML)
+    styles <- readUTF8(stylesXML)
+    styles <- removeHeadTag(styles)
+
+    protection <- getChildlessNode(styles, "protection")
+    wb$styles$hidden <- sapply(protection, function(x) ifelse(grepl("hidden=\"1\"", x), 1, 0))
+    wb$styles$locked <- sapply(protection, function(x) ifelse(grepl("locked=\"1\"", x), 1, 0))
+
```
JanMarvin: to be removed. I'v found a better way to solve this

R/WorkbookClass.R
```diff
@@ -17883,9 +17882,10 @@ Workbook$methods(
           c(getNodes(xml, tagIn = "<font>"), "")
         }))
     } else {
-      fonts <- getNodes(xml = stylesTxt, tagIn = "<font>")
+      fonts <- getNodes(xml = stylesTxt, tagIn = "<fonts")
+      fonts <- getNodes(paste0(fonts, ">"), "<font>")
```
JanMarvin: the original line might have been enough, but I like this one better.

R/WorkbookClass.R
```diff
@@ -17916,7 +17916,7 @@ Workbook$methods(
           regexpr('[a-zA-Z]+(?=\\=".*?")', xfAttrs, perl = TRUE)
         )
       })
-    xfVals <-
+    xfVals <<-
```
JanMarvin: debugging

JanMarvin:
I feel as if I've starred into the abyss and now I'm no longer sure, if the abyss is staring back at me. There are a few things going on in the worksheet linked in #81. The main issue is, that the file itself provides sheets with comments. Currently `openxlsx` is not able to treat these in a way understandable to Excel (but Libre Office and Google Sheets do not complain :shrug: ). I assume that this is partly caused by a bug in `openxlsx` and partly by a misunderstanding of how the xlsx files are supposed to work. Apply this

```diff
diff --git a/R/WorkbookClass.R b/R/WorkbookClass.R
index fb96bde..d4ec13b 100644
--- a/R/WorkbookClass.R
+++ b/R/WorkbookClass.R
@@ -672,7 +672,7 @@ Workbook$methods(
           worksheets_rels[[i]] <<- unique(c(
             worksheets_rels[[i]],
             sprintf(
-              '<Relationship Id="rIdcomment" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/comments" Target="../%s"/>',
+              '<Relationship Id="rId3" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/comments" Target="../%s"/>',
               fn
             )
           ))
diff --git a/R/baseXML.R b/R/baseXML.R
index 812f99a..190d6f1 100644
--- a/R/baseXML.R
+++ b/R/baseXML.R
@@ -201,8 +201,8 @@ genBaseWorkbook <- function() {
 genBaseSheetRels <- function(sheetInd) {
   c(
     sprintf('<Relationship Id="rId1" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/drawing" Target="../drawings/drawing%s.xml"/>', sheetInd),
-    sprintf('<Relationship Id="rId2" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/printerSettings" Target="../printerSettings/printerSettings%s.bin"/>', sheetInd),
-    sprintf('<Relationship Id="rIdvml" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/vmlDrawing" Target="../drawings/vmlDrawing%s.vml"/>', sheetInd)
+    # sprintf('<Relationship Id="rId2" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/printerSettings" Target="../printerSettings/printerSettings%s.bin"/>', sheetInd),
+    sprintf('<Relationship Id="Id2" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/vmlDrawing" Target="../drawings/vmlDrawing%s.vml"/>', sheetInd)
   )
 }
```

and delete the sheets containing the comments.

```R
wb <- openxlsx::loadWorkbook("~/Workbook.Template.new.xlsx")
wb$deleteWorksheet("Lookups")
wb$deleteWorksheet("Basic inputs")
openxlsx::saveWorkbook(wb, file="/tmp/test.xlsx", overwrite=T)
```

And you get a file that works without complains (but lacks two sheets :laughing: ).

Sadly there is more. The current xml parser is not good enough to parse `styles.xml`

* which contains many alignment and protect nodes. The reason it fails is due to some nodes being single line and some multi line. Ultimately the resulting file will be unusable, since certain cells that are supposed to be writable, will be protected.
* `openxlsx` ignores the majority of the font information in this sheet (located in `styles.xml` too). Therefore the file will not only be unusable, but appears visually broken too.

This pull requests fixes both issues. I have solved this by including pugixml a C++ xml parsing library, which is able to read and parse xml files and to return nodes as well as the attributes to the user. It might be possible to fix the `openxlsx`parse some other way, but having an unreliable parser is why there is so much regex stuff going on in the first place.

*STILL* this is not yet enough to write a file that will work flawlessly with Excel. It will still complain about some error when loading, but will finish afterwards. Though it will remove all the the comments.

Back to my initial statement, I have starred at the various xml files, but they simply do not make any sense  to me. I'm not sure what the included printer binaries are doing, nor why Excel complains and to be honest, I'm not even sure that I want to know anymore :smile: . I'll let this rest for the week and see if I'm still interested next week. I would not lose any sleep, knowing that complexe xlsx files as the linked one, simply do not work.

JMPivette:
@JanMarvin I really like the fact that you are trying to use a reliable C++ xml parser to get rid of all these regex and to avoid creation of "broken" workbooks.
Let me know if you need any help. I think these changes will be valuable for all openxlsx users 👍

JanMarvin:
@JMPivette thanks, though before starting a wider inclusion I would like a word from @ycphs or others if he/they is/are considering to go this route. Bundling and replacing functions is definitely a larger task and unless consensus is reached, that this is considered, I will not further go this route. Until now this PR is just a work in progress and my little playground :smile: 

Especially since there are other alternatives which would allow to ditch the `pugixml` inclusion (and might be better maintainable in the future), e.g. R packages `XML` and `xml2` or more fixes for the current reader. A very real issue is that replacing the current functions will introduce new issues, for the mere benefit of having nicer styles. Therefore this PR should not be seen as "do or die" proposal. Similar like I've said before, I could live with the current state of `openxlsx` not fixing anything at all.

JanMarvin:
I've hacked my way into being able to write a xlsx file that contains the comments. Android Excel will still complain, but at least the comments are visible. The issue is with the drawings xml+vml. These files are used for drawing the comment box.

Excel detects that the file was not written by itself (maybe it tries do identify certain pages by uris), therefore the drawing code does not work. In the worst case, excel simply dies on startup. Therefore my initial approach, where I renamed `Idvml` to `rId2` succeeded. Unknowingly I created a situation, where Excel could no longer identify that the drawing code was supposed to be attached to `rId2`. It "got confused" and allowed the user to remove this content on startup.

What I have done: Open the original file in Libre Office and save it (this file can be opened by Android Excel). Hack into `openxlsx` to replace drawing xml+vml with the corresponding Libre Office files. Save the file.

I've begun fleshing out everything unnecessary for reading the file. Printer binaries, extension fields, theme.

```R
# open and save the original file in Libre Office
unzip("~/Workbook.Template.new_LO.xlsx", exdir = "/tmp/lo")

wb <- openxlsx::loadWorkbook("~/Workbook.Template.new.xlsx")

fls <- c("/tmp/lo/xl/drawings/drawing1.xml",
         "/tmp/lo/xl/drawings/drawing2.xml")

for (i in seq_along(fls)) {
  wb$drawings[i] <- openxlsx:::readXML(fls[i])
}

fls <- c("/tmp/lo/xl/drawings/vmlDrawing1.vml",
         "/tmp/lo/xl/drawings/vmlDrawing2.vml")

for (i in seq_along(fls)) {
  wb$drawings_vml[[i]] <- openxlsx:::readXML(fls[i])
}

openxlsx::saveWorkbook(wb, file="/tmp/test1.xlsx", overwrite=T)
unzip("/tmp/test1.xlsx", exdir = "/tmp/xlsx")
```

JanMarvin:
I'm now able to read and save a file that looks identical in LibreOffice (but fails to load on Android Excel, works with the LibreOffice Workaround, but Excel it is still complaining).
There are various things going on, and even though I've removed or commented a lot of the older code, I'm not yet sure how to proceed. The issue when mimicking the original style, is that the ids `openxlsx` reads, are not necessarily identical to the ones Excel has written (or to be precise: I have not found any such case). The rationale behind this might have been obvious in early stages of development and with xlsx files containing only minor "decoration", meaning hidden rows or cols, various fonts, fills and borders. For now I tried to keep the Excel written ids, since that was easier when working with the nodes attributes, otherwise one would have to change these ids to the new ids as well (aaaand I do not know how the new ids are calculated).

Next steps might be to look into creation of drawings similar to LO, add to add calcSheet (I already had that, but did not know what to do with it, LO does not use it). Remove a bunch of hacks, maybe try to come up with a better reference holding mechanism. Re-add functionality and clean this mess up :smile:

ycphs:
I think it is a good idea to use the library. The only downside ist the additional dependency, but I think it is definitely worth for preventing this package to reinvent the wheel.

@JanMarvin thank you for the work on this PR and the development of the package.

JanMarvin:
There is a *lot* going on in this little branch. Basically: I've take `openxlsx` apart and have successfully broken everything and now try to rebuild it piece by piece.

I've written myself a few pugi files, capable of reading and writing xml files. I've included these in various places and am now able to read and write a first xlsx file I'm able to import in LibreOffice. Only numbers, col and row formats, but eh, better than nothing!

Previously everything was imported as single string and handled if needed. Some values were never handled at all, therefore e.g. the file in the issue underlying this pull request could never been imported and written identically. Attributes on cols and rows were not imported and simply assumed to be of a certain value and type. My approach is a bit different as you can see. Having a reliable xml parser at hand I take the files and import nodes and attributes as named character vectors nested in (nested) lists. At least that is how I currently handle it. 
Not every value has to be treated this way, but it allows more flexibility e.g. now you could write a function to hide a previously un-hidden row. Granted it's slower, but then again, I'm currently handling workbooks and there might still be other speed benefits along the way. As in don't read and treat everything prior to writing.

I'm still thinking about the best approach to store the imported values as the current approach is in early stages. I'm still getting familiar with pugi and the overall codebase (S4 classes are not my favorites), some values are still imported twice, there is debug output etc. and tests are expected to fall for the foreseeable future too. [edit] For clarification, currently both, the old and the new parser run on every worksheet (first the old one and finally the new one, replacing various values), which of course is costly. once I've decided which way to go, the old will be deactivated and can be replaced with new functions. [/edit] I'm taking the rest of the week off, if you're willing to have a look, please go ahead.

JanMarvin:
Finally a bit of progress. I'm able to read and write the file from the initial issue ... and it opens in Android Excel! (Excel complains, but hey I'm not greedy :smile: )

It is still a long way though as I'm _currently_ actively avoiding most of what `openxlsx` internally does, which is required for every interaction aside from importing and saving a file without modification. Though, through the extensive use of S4 classes in `openxlsx` functions are often weirdly connected which makes even minor changes quite complicated. (An example: when importing a workbook, it already defines which function can be used to save the content. Unaware of this I spent some time with the code, clueless why it sometimes worked (when I imported the workbook) and sometimes didn't (when I tried to use an imported workbook).)

But then again, I'm quite happy with xml parser as it allowed me to replace the old c++ saving functions (there are _two_ for saving) with rather simple pugixml functions (the code needs cleaning too, I'm aware of this :wink:). I'm looking forward to replacing the old import functions soon and in the long run all of the regex code becoming obsolete.

@JMPivette glad if you could have a first look to see if it compiles on Mac and maybe run a few loadWorkbook/saveWorkbook tests. Aside from that, data is currently stored here:  `wb$worksheets[[x]]$sheet_data$vtyp` so do not expect changes to be saved to the file, even if some of the old functions will work without complains. For now it's merely a proof of concept.

JanMarvin:
Just as reference: data is imported similar to how it's found in the xlsx file. There is a list of length nrow. each row can contain a character vector or list of size ncol. Be aware that ncol can vary for each row (and iirc a row must not begin in the first cell e.g. "A1").

Column attributes r, s and t are stored as characters per row. Column nodes f and v are stored as lists per row. Rows have their own style attributes and must not contain nodes just as columns must not contain nodes.

This differs from the current approach how data is imported. I think that storing data the same way Excel does is beneficial, for instance if one wants to insert/delete a row (insert it between and rename all rows afterwards; oh and ofc all reference in formulas, shared strings etc.), but it might be tricky to insert data in specific cell ranges. One has to check if the cell is already in the rows and if we have to move/modify any or all lists around? Though this problem is bigger for worksheets when we're updating a range, not for newly created sheets or modifications of single cells. Those should be fairly easy.

Another approach I have been considering: get the dimensions and create a data frame like cube. Now add information on all rows, cols, styles, formulas and values and fill the rest with blanks. this will be more memory consuming, but overall easier to update. After all it's just multiple data frames. Then again, Excel has combined rows and cols and many other nasty things and maybe I don't want to go down that rabbit hole. Finally the data should be able to be written into Excel files and therefore this large cube must be searched for valuable information.

JanMarvin:
Function to get a data.frame from an imported workbook

```R
library(openxlsx)

wb2 <- loadWorkbook("/tmp/test.xlsx")
# wb <- loadWorkbook("inst/extdata/loadExample.xlsx")


as_typ <- function(x, types) {
  for(i in seq_along(x)) {
    if(types[i] == "character")
      x[[i]] <- as.character(x[[i]])
    if(types[i] == "integer")
      x[[i]] <- as.integer(x[[i]])
    if(types[i] == "numeric")
      x[[i]] <- as.numeric(x[[i]])
    # dates
  }
  x
}

get_sheet_values <- function(wb, sheetid, types) {
  
  # row, style and type
  rtyp <- wb$worksheets[[sheetid]]$sheet_data$rtyp
  styp <- wb$worksheets[[sheetid]]$sheet_data$styp
  ttyp <- wb$worksheets[[sheetid]]$sheet_data$ttyp
  
  
  vval <- wb$worksheets[[sheetid]]$sheet_data$vval
  
  if (missing(types)) {
    ncols <- max(sapply(rtyp, length))
    types <- rep("character", ncols)
  }
  
  data <- data.frame(
    do.call("rbind", lapply(vval, as.character))
  )
  
  data <- as_typ(data, types)
  
  rtyp <- do.call("rbind", lapply(rtyp, as.character))
  styp <- do.call("rbind", lapply(styp, as.character))
  ttyp <- do.call("rbind", lapply(ttyp, as.character))
  
  attr(data, "rtyp") <- rtyp
  attr(data, "styp") <- styp
  attr(data, "ttyp") <- ttyp
  
  data
}

sd <- get_sheet_values(wb, 1, types = rep("numeric", 11))
attributes(sd)
```

Function to write mtcars to xlsx.

```R
wb <- createWorkbook()

addWorksheet(wb, "Test")

writeData2 <-function(wb, sheet, data) {
  
  
  sheetno <- wb$validateSheet(sheet)
  
  
  sheet_data <- list()
  
  rows_attr <- cols_attr <- vtyp <- vector("list", nrow(data))
  
  cn <- function(x){
    x <- character(0)
    attr(x, "names") <- character(0)
    x
  }
  
  
  # cols_attr <- lapply(seq_len(nrow(data)),
  #                     function(x) c("min" = as.character(min(data[x,])),
  #                                   "max" = as.character(max(data[x,]))))
  cols_attr <- lapply(seq_len(nrow(data)),
                      function(x) c(collapsed="false",
                                    customWidth="true",
                                    hidden="false",
                                    outlineLevel="0",
                                    max="121",
                                    min="1",
                                    style="0",
                                    width="9.14"))
  
  wb$worksheets[[sheetno]]$cols_attr <- cols_attr
  
  
  
  rows_attr <- lapply(seq_len(nrow(data)),
                      function(x) c("r" = as.character(x),
                                    "spans" = paste0("1:", ncol(data)),
                                    "x14ac:dyDescent"="0.25"))
  
  wb$worksheets[[sheetno]]$rows_attr <- rows_attr
  
  # fval list per row
  fval <- lapply(seq_len(nrow(data)),
                 function(x) 
                   lapply(seq_len(ncol(data)),
                          function(x)
                            character(0)))
  
  wb$worksheets[[sheetno]]$sheet_data$fval <- fval
  
  # ftyp list per row
  ftyp <-  lapply(seq_len(nrow(data)),
                  function(x) 
                    lapply(seq_len(ncol(data)),
                           function(x) 
                                      list()))
  
  wb$worksheets[[sheetno]]$sheet_data$ftyp <- ftyp
  
  # rtyp character vector per row 
  # list(c("A1, ..., "k1"), ...,  c("An", ..., "kn"))
  rtyp <- lapply(seq_len(nrow(data)), function(x) 
    paste0(openxlsx:::convert_to_excel_ref(seq_along(data[x,]), LETTERS), x))
  
  wb$worksheets[[sheetno]]$sheet_data$rtyp <- rtyp
  
  # styp character vector per row (style typ, choose 1)
  styp <- lapply(seq_len(nrow(data)), function(x)
    rep("", ncol(data)))
  
  
  wb$worksheets[[sheetno]]$sheet_data$styp <- styp
  
  # ttyp character vector per row ("s", "str" or "" )
  ttyp <- lapply(seq_len(nrow(data)), function(x)
    rep("", ncol(data)))
  
  wb$worksheets[[sheetno]]$sheet_data$ttyp <- ttyp
  
  
  vval <- lapply(seq_len(nrow(data)),
                 function(x) lapply(seq_len(ncol(data)),
                                    function(i) as.character(data[x, i])))
  
  wb$worksheets[[sheetno]]$sheet_data$vval <- vval
  
  
  vtyp <- lapply(seq_len(nrow(data)),
                 function(x) 
                   lapply(seq_len(ncol(data)),
                          function(x) 
                            lapply(1,
                                   function(x)
                                     cn())))
  
  wb$worksheets[[sheetno]]$sheet_data$vtyp <- vtyp
  
  wb$worksheets[[sheetno]]$dimension <- paste0("<dimension ref=\"A1:",
                                               openxlsx:::convert_to_excel_ref(ncol(data), LETTERS),
                                               nrow(data), "\"/>")
  
  # update a few styles informations
  wb$styles$numFmts <- character(0)
  wb$styles$cellXfs <- "<xf numFmtId=\"0\" fontId=\"0\" fillId=\"0\" borderId=\"0\" xfId=\"0\"/>"
  wb$styles$dxfs <- character(0)
  
  wb 
}

# wb$styles <- wb2$styles
writeData2(wb, "Test", data = mtcars)

# wb$worksheets[[1]]$sheet_data$vtyp
# wb$worksheets[[1]]$sheet_data$vval

# writeData(wb, "Test", mtcars, colNames = F, rowNames = F)

saveWorkbook(wb, "/tmp/test_self.xlsx", overwrite = TRUE)
```

JanMarvin:
Everything should be in place that workbooks can be imported. Still some functions are disabled/commented so that importing is not identical to current master. Writing files is possible, too. For imported files this should work well and drastically improve the layout of the imported files as a bunch of new style information can be written as imported. (This includes styles for rows, cols and cells in general.) Overall this allows for changing layouts from R too, one would simply have to modify the respective worksheet style.

As shown in the function above it should be possible to write data directly from R, too. Similar it would be possible to update cells of imported workbooks. But until this works identical to current master - a bunch of further painful changes are required.

For the future:
In theory it should be possible to add rows or cols, insert a col/row in one of the in the many lists and update all row names EVERYWHERE accordingly (yeah, this is why Excel is slow too). Similar other feature requests or ideas such as importing as pre-specified type, importing ranges, updating named regions, import without formula fields (blindly importing them is f*cking dangerous) etc. should be possible.

JanMarvin:
since everyone is kinda silent, I'm just adding more ideas, mostly to get them out of my head. I'm planning on returning a named list from pugi, with the Excel col letters. once this in place we can use them to construct a data frame from a lists of varying lengths combining via their names. this data frame could later be folded back into the structure xlsx expects.

JMPivette:
Hi @JanMarvin 
I tried to reproduce bug #81 with this version. (opening/modifying and saving `Workbook Template new.xlsx`)
I am able to open the file with Excel for Mac. I am just getting an error saying that the file needs to be repaired (something to do with /xl/styles.xml).

I have been quite busy lately and didn't have time to look at the changes you made on this PR. I will try to have a look soon.

JanMarvin:
Thanks for checking! For now this is expected. I'm not entirely sure why, but hopefully it will reveal itself. If your Excel mentions styles.xls that's a hint I'll look into. Android Excel sadly doesn't tell me anything

JanMarvin:
@JMPivette Thanks to your hint I implemented a workaround. This imports the original `styles.xml` file and the warning - when opening the file in Excel - goes away. We should figure out, what is missing from styles when importing this file. (The last nodes are missing and some attributes, but the file was to large to spot all differences and a diff did not help.)

On a side note, the import returns a list with a named list for now. Look something like this:
```R
> wb$worksheets[[1]]$sheet_data$vval[[1]][1:3]
$F
character(0)

$G
[1] "3579"

$H
character(0)
```

The data can be obtained like this:
```R
lst <- wb$worksheets[[1]]$sheet_data$vval

dt <- data.table::rbindlist(lst, fill = TRUE)
data.table::setDF(dt)

df[names(df)[order(openxlsx::convertFromExcelRef(names(df)))]]
```

ycphs:
If it helps, I can offer to support the changes in the checks. @JanMarvin do you think it's already possible to adapt the checks or should we wait until the implementation of pugixml for the functions is completed?

JanMarvin:
Hi @ycphs, 

thanks, but my implementation is not yet at a stage where I could think about updating checks. Although it has made good progress, it's still a mammoth task to finalize it. An awful lot of functions have to be rewritten. The entire pull request has kind of evolved into a rewrite into `openxlsx 2.0`.
Never the less it is something that would improve `openxlsx` in so many ways. If I watch the issue list, I assume that 90% of the current issues could be closed (at least all these XML/style related like #81 or suffering from nasty regexes like #142). But implementing takes time. One has to understand the old code, think of something new.

Major task on my todo list is implementing the saving/modifying functions e.g. everything that allows writing data.frames from R to Excel. Once it is implemented in a non hacky way, the ~~rest~~ remaining things should fall together a bit easier and provide the flexibility, to do more changes on the fly.
Sadly time is something I'm rather short of currently (blame my everyday job :wink: ). Therefore I cannot give any ETA (but my feeling is that it is more a question of months than days or weeks). Until then any help would be appreciated. Reviews of my changes or suggestions or contributions to this pull request. Currently you cannot really break anything, as most things are broken anyways :smile:

JanMarvin:
Pushed the polished function from above to write data from R. This basically creates the data structure my export function expects and allows writing R objects of type data frame, matrix and vector to positions inside of Excel worksheets. This remains a dummy function and its functionality should be integrated into the already existing `openxlsx` writeData function.

It does not handle dates, formulas, shared strings, updating or inserting into sheets or any style related functions. This is just to write numeric and character data from R to xlsx files.

```R
library(openxlsx)

# create a workbook and add some sheets
wb <- createWorkbook()

addWorksheet(wb, "sheet1")
writeData2(wb, "sheet1", mtcars, colNames = TRUE, rowNames = TRUE)

addWorksheet(wb, "sheet2")
writeData2(wb, "sheet2", cars, colNames = FALSE)

addWorksheet(wb, "sheet3")
writeData2(wb, "sheet3", letters)

addWorksheet(wb, "sheet4")
writeData2(wb, "sheet4", as.data.frame(Titanic), startRow = 2, startCol = 2)

saveWorkbook(wb, file = "/tmp/test.xlsx", overwrite = TRUE)

# load an existing workbook (#81) and add a new sheet to it
wb2 <- loadWorkbook("~/Workbook.Template.new.xlsx")

addWorksheet(wb2, "new_sheet")
writeData2(wb2, "new_sheet", mtcars)

saveWorkbook(wb2, file = "/tmp/test2.xlsx", overwrite = TRUE)
```

JanMarvin:
The last commit reads `inlineStr`. Using the file from #112. 
This can be used to create the correct dataset. See example code down below. Thanks to @sjewo helping me with the merge :smiley: 


```
> wb <- loadWorkbook("~/gh_112.xlsx")
> ilst <- wb$worksheets[[1]]$sheet_data$isval
> vlst <- wb$worksheets[[1]]$sheet_data$vval
> 
> idt <- data.table::rbindlist(ilst, fill = TRUE, idcol = "row")
> vdt <- data.table::rbindlist(vlst, fill = TRUE, idcol = "row")
> 
> cd <- idt[, sapply(idt, function(x) !all(is.na(x[-1]))), with=F]
> nd <- vdt[, sapply(vdt, function(x) !all(is.na(x))), with=F]
> namd <- idt[1, ]
> 
> finaldata <- merge(cd, nd, by="row")
> finaldata[ , row := NULL]
> data.table::setcolorder(finaldata, order(names(finaldata)))
> data.table::setnames(finaldata, old = names(idt), new = as.character(namd), skip_absent = TRUE)
> finaldata
   ​PairIndex Drug1 Drug2 Conc1 Conc2            Response ConcUnit
​1:         1   abc   def 10000 10000  -1.796071094480823       nM
​2:         1   abc   def 10000  3000   1.010289990645463       nM
​3:         1   abc   def 10000  1000  0.4490177736202058       nM
​4:         1   abc   def 10000   300                   0       nM
​5:         1   abc   def 10000   100  0.8980355472404116       nM
​6:         1   abc   def 10000    30  0.1122544434050514       nM
​7:         1   abc   def 10000    10   3.143124415341441       nM
​8:         1   abc   def 10000     0   1.459307764265669       nM
​9:         2   abc   def 10000     0   1.683816651075772       nM
10:        2   abc   def 10000    10 -0.7857811038353602       nM
```

JanMarvin:
The latest commit implements minimal updates to imported xlsx, leaving the style intact. This can be shown using the file from #81. The following code updates two cells of the original file, adding text to `G1` and a number to `H1` on sheet `Lookup` (iirc this was the initial issue) .

I'm thinking about implementing something like this for data frames and matrix objects (updating multiple cells at once) and similar modifications for cells not yet existing in the original data set (requires moving all the other cells a bit around and maybe updating the entire range.

```r
fl <- "~/Workbook.Template.new.xlsx"
wb <- openxlsx::loadWorkbook(fl)
wb <- update_cell(x = "Updated 2021-04-02", wb = wb, sheet = "Lookups", cell = "G1")
saveWorkbook(wb, file = "/tmp/test1.xlsx", overwrite = TRUE)
wb <- update_cell(x = 7, wb = wb, sheet = "Lookups", cell = "H1")
saveWorkbook(wb, file = "/tmp/test2.xlsx", overwrite = TRUE)
unzip("/tmp/test2.xlsx", exdir = "/tmp/xlsx")
```

Note to self: `git rebase -Xtheirs master`

ycphs:
I will try to implement the newly added features with the new parser.

- [ ]  `write.xlsx()` can now handle `colWidths` passed as either a single element or a `list()`
- [ ]  Added ability to change positioning of summary columns and rows.
      These can be set with the `summaryCol` and `summaryRow` arguments in `pageSetup()`.
- [ ]  activeSheet allows to set and get the active (displayed) sheet of a worbook.
- [ ] setting of global options

JanMarvin:
thanks, I think this pull request is still lacking things, but it's a good idea to get familiar with the code base (e.g. currently only `saveWorkbook()` is implemented) Sadly I'm currently lacking time to work on this

JanMarvin:
I've implemented a simple dummy function to import a Workbook into a DataFrame. This function was written just to demonstrate the general idea of the implementation. I've added a few options as well. I assume that the next step will be writing from R over a cellrange and exporting that into a readable xlsx file.

```R
> wb_to_df(wb1)
    Var1 Var2 NA  Var3  Var4       Var5         Var6    Var7
2   TRUE    1 NA     1     a 2015-02-07 3209324 This #DIV/0!
3   TRUE   NA NA #NUM!     b 2015-02-06         <NA>    #N/A
4   TRUE    2 NA  1.34     c 2015-02-05         <NA>   #NUM!
5  FALSE    2 NA  <NA> #NUM!       <NA>         <NA>    <NA>
6  FALSE    3 NA  1.56     e       <NA>         <NA>    <NA>
7  FALSE    1 NA   1.7     f 2015-02-02         <NA>    <NA>
8   <NA>   NA NA  <NA>  <NA> 2015-02-01         <NA>    <NA>
9  FALSE    2 NA    23     h 2015-01-31         <NA>    <NA>
10 FALSE    3 NA  67.3     i 2015-01-30         <NA>    <NA>
11  <NA>    1 NA   123  <NA> 2015-01-29         <NA>    <NA>

> wb_to_df(wb2)
   PairIndex Drug1 Drug2 Conc1 Conc2   Response ConcUnit
2          1   abc   def 10000 10000 -1.7960711       nM
3          1   abc   def 10000  3000  1.0102900       nM
4          1   abc   def 10000  1000  0.4490178       nM
5          1   abc   def 10000   300  0.0000000       nM
6          1   abc   def 10000   100  0.8980355       nM
7          1   abc   def 10000    30  0.1122544       nM
8          1   abc   def 10000    10  3.1431244       nM
9          1   abc   def 10000     0  1.4593078       nM
10         2   abc   def 10000     0  1.6838167       nM
11         2   abc   def 10000    10 -0.7857811       nM
```

JanMarvin:
The non-optimized dummy function should be able to act as an drop in replacement for `read.xlsx`. Regarding the options it only lacks `check.names`, `sep.names`, `na.strings` and `fillMergedCells`. All of these are possible, but currently not highly prioritized. Newly implemented features are importing pre-defined dimensions (`"A1:B2"`), returning the formulas instead of the stored values, importing the character type `inlinestr` and specifying the column type while importing.

JanMarvin:
FYI: I think I have made reasonably good progress. After putting in a lot of brain power and sleep deprivation over the weekend, I was able to report success. I have combined and unified the logics to import.

Two large linked lists are now generated. One contains the values of the `<v>`, `<f>` and `<is>` nodes, the other the attributes of the parent `<c>` nodes (at least the cellname `r=`). This contains all the details (pretty much analogous) to the structure of the xml tree. 

My small dummy function for converting Excel to R-Data.frames is already converted to it. Although it is still programmed exclusively in R, it is already more than satisfactorily fast. In addition, the update_cell function has already been adapted and extended so that longer contiguous areas can now also be overwritten. Due to the achieved improvements I am positive that the saveWorkbook function can be adapted soon and also the writing on new cell ranges should not be a special obstacle anymore. 

The linked lists are named after rows and column names of the Excel sheet:

```R
> wb1$worksheets[[1]]$sheet_data$ctyp[["1"]][["B"]]
$r
[1] "B1"

$t
[1] "s"

> wb1$worksheets[[1]]$sheet_data$cval[["1"]][["B"]]
$v
[1] "2097"
```

_It's quite late therefore: Translated with www.DeepL.com/Translator (free version)_

JanMarvin:
Writing is back! It is possible to write all the imported attributes from `<row>` and `<c>` and all values of `<f>`, `<is>` and `<t>`. Attributes to `<f>` are currently unhandled (I had simply forgotten; I might add them in a third list or - better - I might combine everything into a single list where we have something like `"1"$attributes`, `"1"$values`, and `"1"$secondaryttributes`. The latter might be beneficial when slicing and splicing the lists to add or remove cells and rows.). :smile: 

```R
# inlinestr
xlsxFile <- system.file("extdata", "inlinestr.xlsx", package = "openxlsx")
wb2 <- loadWorkbook(xlsxFile)

# read dataset with inlinestr
wb_to_df(wb2)
saveWorkbook(wb2, file = "/tmp/test2.xlsx", overwrite = TRUE)
```

This file was written and could be opened with LibreOffice (still no Excel on my end). How does an `inlineStr` look like in my current data structure? I'm using this structure, because it eases modifications to the data while keeping the file very similar to the original data. If you want to change the value of "A1", simply remove "is" and "t" and add "v" to the cval list. The writer goes through the list, takes the data for every single row and creates the xml-output for this row. The lines are combined and written to the output file.

```R
# ctyp is "inlineStr"
> wb2$worksheets[[1]]$sheet_data$ctyp[["1"]][["A"]]
$r
[1] "A1"

$t
[1] "inlineStr"

# and cval is the text. that's all
> wb2$worksheets[[1]]$sheet_data$cval[["1"]][["A"]]
$is
[1] "PairIndex"
```

JanMarvin:
The handling of function attributes is now added. (There is still space for improvement. I've added it as an attribute that is added to every cell, even though it's currently only available/implemented with formulas. The reason why I'm doing it is, it is easier to access a field and ignore it, than to check if the field is available and filled. C++ has the nasty habbit of crashing when accessing a non available object.) I was able to read the workbook from the original issue and save it. The file opens in Excel as well and everything looks alright. I've restored the dummy functions to read from a workbook into a dataframe and the dummy function to update cells. Last but not least I've updated the `writeData2()` dummy function. It's a wrapping an R object and creates the data structure the writer expects. The handling of the data in c++ has gotten much easier, now that things are more or less in place.

So this works as well:

```R
# create a workbook and add some sheets
wb <- createWorkbook()

addWorksheet(wb, "sheet1")
writeData(wb, "sheet1", mtcars, colNames = TRUE, rowNames = TRUE)

addWorksheet(wb, "sheet2")
writeData(wb, "sheet2", cars, colNames = FALSE)

addWorksheet(wb, "sheet3")
writeData(wb, "sheet3", letters)

addWorksheet(wb, "sheet4")
writeData(wb, "sheet4", as.data.frame(Titanic), startRow = 2, startCol = 2)

saveWorkbook(wb, file = "/tmp/test.xlsx", overwrite = TRUE)
```

The next thing on the to do list is updating the cells that are currently not in the xml worksheet. It should be possible to borrow what the `writeData2()` function does, but the data has to be embedded into the imported list. Once we can import and export the workbook, and read the data to R and modify it as well, we should consider what to do next.

JanMarvin:
This weeks sprint is coming to an end. I've just pushed a dummy implementation how to update a worksheet (not my best work granted, but it worked in my examples). Examples have been added, a very basic one is the following:
```R
# import an xlsx file with inlinestrs
wb <- loadWorkbook(system.file("extdata", "inlinestr.xlsx", package = "openxlsx"))

# write mtcars dataset from D4 on
writeData(wb, 1, mtcars, startCol = 4, startRow = 4)

# save the file, loads in android excel
saveWorkbook(wb, "/tmp/test.xlsx", overwrite = TRUE)

# reimport the range
wb_to_df(wb, dims = "D4:N36")
```

This concludes my dummy implementation of pugixml as reader and writer. Sadly that's not the end of the story. There is still *a lot* broken. Lets face it, I've more or less taken the entire xml loading process, taken it apart and rewritten it in the nasty S4 classes. Therefore many things are still missing or incomplete with my approach (for example: `openxlsx` adds its own r:ids all over the place (for god knows whatever reason), therefore drawings and images are misplaced or missing). The `wb_to_df` and the `update_cell` part of the implementation are almost exclusively written in R and not optimized (therefore speed varies between fast and slow as hell). [edit] In the write date step[/edit] I only handle `numerics` and `characters`, not as shared strings (`sst`) just as basic `inlineStr`. No dates, no formulas, nothing else is implemented. To be honest anything that is still working, is probably simply not yet broken.

As I've stated before, the entire process has begun to evolve into another project, but the good news (at least for me) is, many things are possible and it should be rather easy to implement them (it's mostly a matter of wrangling the data into the lists as shown in the dummy implementation). We need formula attributes to identify array formulas? Just add it. We want to hide a row? Just add the attribute. We want to modify a certain style attribute? Just do it. I like simple and I like how nice the pugi implementation is. Honestly I consider the loading and writing functions really beautiful code :wink:.

I've enabled the `read.xlsx` and `readWorkbook` functions. If you wanna try things out or see for yourself how well some of the code behaves with various files, now would be an _excellent_ moment. As usually comments to the entire endeavor are welcome too.

JanMarvin:
I'd be interested, how the implementation performs regarding time and memory consumption with real world examples. Everyone knows Excel is no database, but still there are some wild sheets out there. Therefore I'd like to know how people use `opexlsx`. Maybe accuracy in every detail is preferred to being super fast, but maybe speed in reading dataframes to R is everything because there are hundreds of files and you don't want to wait. Just let me know.

I've done some microbenchmarks on the read example (readTest.xlsx iirc, the one with the 8k lines of mostly empty cells) and they've shown that the pugixml implementation is way slower (like 14 milliseconds vs 700). But this is expected: just like Excel in the real world, the pugixml function imports every attribute of every row and cell in the sheet and does not care if it's meaningful or not. Excel Files with thousands of formatted cells and rows without values are large too and loading speed is bad. Believe me, I've seen it.

Still, there are possible ways to change this behavior and speed things up a bit.

* read only cells with a value, e.g. must have `<v>`, `<is>` or `<f>` node
* read only pre selected attributes
* load only a selected sheet
* read only in preselected dimensions, e.g. "A1:B4"
* ...

Especially, when importing only to print things to R, a faster slimmed approach seems reasonable. But that's some extra work and nothing has been done in that regard. For now, I've experimented with calculating the length of every list (no `push_back` anymore), but the improvements where not that huge and I want the implementation to rest for now (not gonna rewrite 5 functions on a daily basis 😊).

JanMarvin:
I'm closing this one. The last push to this contains a working implementation, but for me this has somehow evolved into a project on it's own and most likely it will never be implemented as a single pull request (most definitely not in this messy way). It's slow (due to the heavy usage of Rcpp::Lists in the implementation. Creating and copying data from C++ to my R lists takes forever), but it works and it was a fun learning experience.

ycphs:
@JanMarvin 
I looked into the performance issues and did some testing.

My findings so far
- pugixml is a great choice and easy to implement
- I tested some bigger files and pushed the data during the loops into a c++ vector and combined it at the end to an Rcpp list/data frame. That increases the performance.
- I was even able to read the file with the missing r attributes from issue #227
- if we add data.table as dependency, the reading of the data files and creating of data.frames or data.tables seems for my tests really fast.

I have some questions:
- do you want to create a new project from scratch for implementing pugixml?
- what are your thoughts about adding data.table as dependency and removing some other dependencies which are only use for examples?

JanMarvin:
Hi @ycphs ,

thanks for your reply. I try to comment everything and to answer your final question, the answer is there is no general plan for the future.

* I too like pugixml. I'm always amazed how simple the interface is and how easy the handling of XML files becomes.

* Since I've closed this PR, I've further tweaked the performance by switching to R's C functions instead of Rcpp types list and character vector. Therefore the loadWorkbook comparison with the readTest file is down to a merely 50% increase. That's in the range I've aimed for 🚀 . For smaller files without that many attributes per cell, the pugixml implementation is already faster than the openxlsx reader and for different tasks there is still room for improvement: Reading only selected cells, rows or attributes (just like the current reader does). Still there might be faster ways to do things, export only character vectors or data frames, but I wanted to keep things together and to be able to easily modify them.

* I've also imported the file from #227 successfully. Well at least my tests indicated this.

* Regarding the import functions, I don't think that data.table must be included. My dummy functions are still R code exclusive and rely on a heavy loop. A thing R is pretty slow handeling. (Basically the loop runs over all rows and cols of the data frame and the list. It was written just to test my implementation, to show what is possible and to experiment with.) Once this loop is migrated to Rcpp it should be blazing fast.
In general I'm all for removing unnecessary dependencies!

Regarding your remaining question. I'm not sure how I'm going to proceed. For now I've locally switched to a openxlsx2 repository, but that's just for my development and testing and allows me to keep openxlsx intact. Other than me developing for myself there currently is no greater plan.

Maybe we can have a zoom call sometime in the future and discuss this further (using the technological benefits of a pandemic).

Issue #126, opened 2020-12-14
Error when saving a file with `Threaded comments`


JMPivette:
**Describe the bug**
If we load a Workbook that contains 'Threaded comments' we get an error if we want to save this file

**To Reproduce**
1. use loadWorkbook() on this file
[comment.xlsx](https://github.com/ycphs/openxlsx/files/5688306/comment.xlsx)

2. Try to save the loaded workbook with saveWorkbook()

3. We get the following error:
`Error in comment_list[[i]]$style[[j]] : subscript out of bounds`

**Desktop (please complete the following information):**
 - R [4.0.2] 
 - Version [4.2.3]

**Additional context**
It looks like openxlsx is not compatible with the recent "Threaded Comments" from Excel and works only with the classical comments (also names "Notes").

The error occurs when saveWorkbook() tries to add a "style" information to the XML. This information doesn't exist in "Threaded Comments"

JMPivette:
Linked to the following bugs:
https://github.com/awalker89/openxlsx/issues/486

JMPivette:
Fixed by PR #135

PR #133, opened 2021-01-02
Update openxmlformats. Closes #74

JanMarvin:
This pull request fixes #74 and improves #81.

Due to some outdated or not entirely included openxmlformats the file from this issue got corrupted. The openxmlformats were obtained from #81. 

```R
wb <- openxlsx::loadWorkbook("/tmp/R_data_validation_test.xlsx")
openxlsx::saveWorkbook(wb, "/tmp/test.xlsx", overwrite=T)
```

PR #135, opened 2021-01-07
Fix issue #126 

JMPivette:
I started to work on this as a "simple" problem on comment saving when no styling was provided (#126).

In the end it was much more complicated and I had to add `threadComments ` and `persons` information to the WorkBook object. So now "Threaded Comments" are preserved when saving a loaded workbook.

A Unit test has been added corresponding to bug 126.
Manual tests have also been done on several big xlsx files.

@JanMarvin , During my test I tried to load and save the file from bug #81 and it worked without corrupting the output file. You can have a look but it might have solved this bug as well. 🤞 

A review of the code is welcome. I understand a bit more how the package works but I hope I didn't break anything 😄


JanMarvin:
Awesome @JMPivette , thanks for bringing this up. I'll have a look later!

[Edit:] I did a quick run, but for me the file still appears to be broken. I run this on Arch Linux and on Window 10 and tested it with Excel on Android.

```R
# download file
tmp <- tempdir()
tmpxlsx <- paste0(tmp, "/Workbook_Template_New.xlsx")

curl::curl_download("https://github.com/ycphs/openxlsx/files/4879159/Workbook.Template.new.xlsx",
                    tmpxlsx)

# load it and save it 
wb <- loadWorkbook(tmpxlsx)
saveWorkbook(wb, "/tmp/test.xlsx", overwrite = TRUE)
```

R/WorkbookClass.R
```diff
@@ -24,6 +24,7 @@ Workbook$methods(
         category = category
       )
     comments <<- list()
+    threadComment <<- list()
```
JanMarvin: Maybe `threadComments`?

R/WorkbookClass.R
```diff
@@ -585,6 +590,8 @@ Workbook$methods(
     nPivots <- length(pivotDefinitions)
     nSlicers <- length(slicers)
     nComments <- sum(sapply(comments, length) > 0)
+    nThreadComment <- sum(sapply(threadComment, length) > 0)
```
JanMarvin: Similar `nThreadComments`?

R/WorkbookClass.R
```diff
@@ -686,6 +693,46 @@ Workbook$methods(
 
       .self$writeDrawingVML(xldrawingsDir)
     }
+    
+    ## Threaded Comments xl/threadedComments/threadedComment.xml
+    if (nThreadComment > 0){
+      xlThreadComment <- file.path(tmpDir, "xl", "threadedComments")
```
JanMarvin: Maybe `xlThreadedComments`

R/WorkbookClass.R
```diff
@@ -686,6 +693,46 @@ Workbook$methods(
...
+      dir.create(path = xlThreadComment, recursive = TRUE)
+      
+      for (i in 1:nSheets) {
```
JanMarvin: Maybe `seq_len(nSheets)` (but not sure if that works here)

R/WorkbookClass.R
```diff
@@ -686,6 +693,46 @@ Workbook$methods(
...
+            from = fl,
+            to = file.path(xlThreadComment, basename(fl)),
+            overwrite = TRUE,
+            copy.date = TRUE
+          )
+
+          worksheets_rels[[i]] <<- unique(c(
+            worksheets_rels[[i]],
+            sprintf(
+              '<Relationship Id="rIdthread" Type="http://schemas.microsoft.com/office/2017/10/relationships/threadedComment" Target="../threadedComments/%s"/>',
```
JanMarvin, line 715: I'm not entirely convinced that not providing a precise id is a working solution, as I suspect this to be a possible cause of corruption (might work though). It's like this for comments and for vml, so :shrug:

R/loadWorkbook.R
```diff
@@ -70,6 +69,8 @@ loadWorkbook <- function(file, xlsxFile = NULL, isUnzipped = FALSE) {
   vmlDrawingXML <- xmlFiles[grepl("drawings/vmlDrawing[0-9]+\\.vml$", xmlFiles, perl = TRUE)]
   vmlDrawingRelsXML <- xmlFiles[grepl("vmlDrawing[0-9]+.vml.rels$", xmlFiles, perl = TRUE)]
   commentsXML <- xmlFiles[grepl("xl/comments[0-9]+\\.xml", xmlFiles, perl = TRUE)]
+  threadCommentXML <- xmlFiles[grepl("xl/threadedComments/threadedComment[0-9]+\\.xml", xmlFiles, perl = TRUE)]
```
JanMarvin: Maybe: `threadedCommmentsXML`

tests/testthat/test-loading_workbook.R
```diff
@@ -923,6 +923,14 @@ test_that("Loading multiple pivot tables: loadPivotTables.xlsx works",{
...
                2)
-  
+})
+
+test_that("Load and saving a file with Threaded Comments works", {
+  ## loadThreadComment.xlsx is a simple xlsx file that uses Threaded Comment.
+  fl <- system.file("extdata", "loadThreadComment.xlsx", package = "openxlsx")
+  wb <- loadWorkbook(fl)
+  # Check that wb can be saved without error
+  expect_error(saveWorkbook(wb, file = tempfile()),
```
JanMarvin: doesn't this expect an error? Maybe `expect_silent` or `expect_output_file`?

JanMarvin (approved):
Looking good! Haven't seen a threaded comment in the wild, but an improvement is an improvement. The test is kinda wrong I guess, my other comments are only style related.

tests/testthat/test-loading_workbook.R
JMPivette: In fact there is no function `expect_no_error()` in `testthat` and one way to test it is to use `expect_error()` with argument `regexp = NA`.

From `expect_error()` documentation:

> [...] If NA, asserts that there should be no errors.[...]

But you are right `expect_silent()` is probably a more readable option here.

JanMarvin: Did not know that!

R/WorkbookClass.R
JMPivette: I had exactly the same concerns but decided to use the same logic as `comments` because it was easier 😄 
I agree, it is not really clean and might need to be cleaned along with comments and vml.

From what I have seen Microsoft Office only complains if there are duplicated `Relationship Id`in an xml file. It doesn't seems to care about the value itself.

JMPivette:
> [Edit:] I did a quick run, but for me the file still appears to be broken. I run this on Arch Linux and on Window 10 and tested it with Excel on Android.
> 
> ```r
> # download file
> tmp <- tempdir()
> tmpxlsx <- paste0(tmp, "/Workbook_Template_New.xlsx")
> 
> curl::curl_download("https://github.com/ycphs/openxlsx/files/4879159/Workbook.Template.new.xlsx",
>                     tmpxlsx)
> 
> # load it and save it 
> wb <- loadWorkbook(tmpxlsx)
> saveWorkbook(wb, "/tmp/test.xlsx", overwrite = TRUE)
> ```

I did some additional test and these lines of code work on my computer and I can open the file with Excel on MacOS.
The problem hasn't been fixed by this PR but by the one made previously: PR #133

Issue #80, opened 2020-06-24
Dates older than 1 March 1900 are incorrectly converted to/from Excel


FrancoisR95:
**Describe the bug**
Dates older than 1 March 1900 are incorrectly converted to/from Excel because openxlsx is unaware that Excel includes a non-existing 29 Feb 1900 day (see https://docs.microsoft.com/en-us/office/troubleshoot/excel/wrongly-assumes-1900-is-leap-year) which is coded in Excel with the number 60.
This bug happens in both ways:
1. from XLSX to R (e.g. convertToDate function) 
2. from R to XLSX (writeData function)

**1. To reproduce with convertToDate**
<code>convertToDate(1, origin = "1900-01-01")</code>
The result is incorrectly:
<code>[1] "1899-12-31"</code>
The result should be:
<code>[1] "1900-01-01"</code>

**2. To reproduce with writeData**
<pre><code>df <- tribble(~Date, as.Date("1900-01-01"))
tmp_XLSX <- tempfile()
wb1 <- createWorkbook()
addWorksheet(wb1, "Sheet1")
writeData(wb1,"Sheet1", df)
saveWorkbook(wb1, tmp_XLSX)
openXL(tmp_XLSX)</code></pre>
This opens an Excel file with the date <code>1900-01-02</code>

Remark:
writeData converts the date to Excel by adding 25569 to all dates (in case the date origin is 1900) but it should add only 25568 if the date is < "1900-03-01" (i.e. < -25508).

Chris-Stockbridge:
If someone can help me find the writing half of this conversion I would be happy to help fix that also.

JanMarvin:
https://github.com/ycphs/openxlsx/blob/6b45bd92c4ed202fc27abb45c6850d9acfb4d519/R/workbook_write_data.R#L31-L61
But there might be other lines involved too. The handling of workbook and default might be separated (just had a quick look).

PR #153, opened 2021-03-26
fix convertToDate() for dates before 1900-03-01

Chris-Stockbridge:
related to issue #80. This fixes reading only. Writing is not fixed.

This is not an issue for most commonly used dates, but it is a problem for someone testing the function against what should be a known good date such as "1900-01-01" => 1

### DIFF
--- a/R/wrappers.R
+++ b/R/wrappers.R
@@ -2574,8 +2574,10 @@ worksheetOrder <- function(wb) {
 convertToDate <- function(x, origin = "1900-01-01", ...) {
   x <- as.numeric(x)
   notNa <- !is.na(x)
+  earlyDate <- x < 60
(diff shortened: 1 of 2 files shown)


ycphs:
Thank you very much

ycphs:
Could you please provide also some tests?

Chris-Stockbridge:
@ycphs commit 1a90f8a2581c1ca8388e9eccdcc7e0e1886f0e2a provides tests for reading/writing round trip, and 549cb2f750f33f1b0cb4a347b7bde186975f692f provided a test for the convertToDate() function. Is there something specific you want tested also, or did you miss those?

ycphs:
@ChrisSUMich sorry I didn't checked the commits.

all good, sorry for the misunderstanding

PR #156, opened 2021-03-28
Update actions and readme

ycphs:
remove unnecessary badges


ycphs:
ubuntu 20.04 is now the ubuntu-latest, which made the changes necessary to keep the CI working.

PR #157, opened 2021-03-28
Too much data

ycphs:


R/workbook_write_data.R
```diff
@@ -240,6 +227,33 @@ Workbook$methods(writeData = function(df, sheet, startRow, startCol, colNames, c
   newStrs <- v[strFlag]
   if (length(newStrs) > 0) {
     newStrs <- replaceIllegalCharacters(newStrs)
+    vl <- stri_length(newStrs)
+    
+    for (i in which(vl > 32767)) {
+      
+      if(vl[i]>32768+30){
+        message(
```
JanMarvin: maybe even a `warning`? ultimately this could lead to the loss of data, even though it should not occur that often.

JanMarvin: most likely it's super obvious what is going on here, but maybe add a comment what is supposed to happen? e.g. why add 30 than 15 etc. in the first place? Why not something along the lines line 255 and if v and vl differ print a warning?

ycphs: - I will change it to a warning
- I set it up to know a little bit about the truncated string.

@JanMarvin thank you for the review

R/workbook_write_data.R
```diff
@@ -227,6 +227,33 @@ Workbook$methods(writeData = function(df, sheet, startRow, startCol, colNames, c
...
+          paste0(
+            stri_sub(newStrs[i], 32768, -1),
+            " is truncated. 
+Number of characters exeed the limit of 32767."
+          )
+        )
+        
+      }
+      
+      # v[i] <- stri_sub(v[i], 1, 32767)
```
JanMarvin, line 255: isn't something like this required?

ycphs: I had the following in mind:
- for data which is just a bit over the limit it is shows the truncated part
- for data which is way over the limit the warning shows the head and the tail of the truncated data
- this additional information doesn't have a big impact on the performance, but I would like to know which infromation I might not have in my workbook.

R/workbook_write_data.R
epruesse: @JanMarvin @ycphs - Apologies for late response here. 

Excel will ***always*** truncate fields 32kb before showing the file. It will do that only after a "**do you trust this file**" popup, making the user worry about the file being malicious, and a "**do you want to repair this file**" popup, making the user believe `openxlsx` created a broken file.

So for the non-expert, issuing a warning in R and then truncating so Excel will just open the file cleanly is very much desirable. Otherwise the files just look defective at best and malicious at worst. I'd add a `options(openxlsx.max_field_len=2**15-1)` option so that people who know what they are doing can decide to write longer fields without changing the source code.

PR #174, opened 2021-04-12
165 options

### LINKED ISSUE #165: Request:  More global options
Specifically these come to mind for `addWorksheet`, `writeData`, and `writeDataTable`

* showGridLines
* tabColour
* borders
* borderColour
* borderStyle
* tableStyle
* headerStyle
* withFilter
* na.string
* firstColumn
* lastColumn
* bandedRows
* bandedCols

... or essentially anything else that could be used for formatting.

Further, the default options could be saved in an exported `op.openxlsx` for reference.  If that is used, then the default options can be set globally with `.onAttach()` and a wrapper for `getOptions()` specifically for `openxlsx` can be used to simplify each of the calls.  I'd expect that all of the calls to these would have to be reviewed to make sure that the global options set are checked first.

jmbarbone:
Resolves #165 

A few notes:
* `openxlsx_getOp()` is a safe way to retrieve an openxlsx option
  * checks first for the name in `op.openxlsx` first -- this should prevent misspellings (already has for me)
  * a `default` can be assigned for `NULL` values -- which will also make the function defaults visible
* options that have not been set are set `.onAttach()`
* Documentation added for the `op.openxlsx` object so make all options user facing (although more descriptions can probably be added)
* Only one test has been edited but I don't believe I wrote it correctly in the first place.


ycphs:
Thank you very much. 

maybe we could also use package [options](https://stat.ethz.ch/R-manual/R-patched/library/base/html/options.html)

R/openxlsx.R
```diff
@@ -18,30 +18,147 @@
...
+  
+  if (!x %in% names(op.openxlsx)) {
+    warning(
+      x, " is not a standard openxlsx option\nCheck spelling",
+      call. = FALSE
+    )
+  }
+  
+  x
+}
```
jmbarbone, line 164: > Thank you very much.
> 
> maybe we could also use package [options](https://stat.ethz.ch/R-manual/R-patched/library/base/html/options.html)

@ycphs Not quite sure what you mean.  These functions use the `option()` function for setting the global options.  There is an additional step in which `openxlsx_getOp()` and `openxlsx_setOp()` also check that the option is present in the `op.openxlsx` object to validate the name is given appropriately (I had seen a few instances of trying to retrieve options with names like `dateTimeFormat` and `datetimeFormat`.  Users are still free to use either approach:

```r
# explicitly adding the openxlsx. prefix
op1 <- list(openxlsx.gridLines = FALSE, openxlsx.withFilter = FALSE)
op2 <- list(gridLines = FALSE, withFilter = FALSE)

options(op1) # valid and works fine
options(op2) # valid but won't set the openxlsx options

# appends the openxlsx. prefix is not provided, results are the same
openxlsx_setOps(op1)
openxlsx_setOps(op2)

openxlsx_setOps(withfilter = FALSE) # will throw a warning because the name isn't exactly right
```

Another example of this sort of implementation is [in the tibble package](https://github.com/tidyverse/tibble/blob/dd29fb76b6b0335a08658c14483fe3cbb81cf7d7/R/options.R).

PR #175, opened 2021-04-14
Add detail to the error when failing to import dates

### LINKED ISSUE #172: c++ exception when importing a data column with mixed type
A column with mixed date and text types causes a c++ exception.

**To Reproduce**
1. Create an XLSX file with two cells. One filled with a date and set to "Date" display type, the other with text such as '2001'.
2. Attempt to import this file using read.xlsx()
3. See that it fails with an error message of:
`Error in read_workbook(cols_in = cell_cols, rows_in = cell_rows, v = v,  (myScript.R#95): c++ exception (unknown reason)`

**Expected behavior**
Any of:
1) Return an error specifying what happened (possibly including the row/column with the error)
2) Return NA for invalid dates and issue a warning
3) Return the column as text. Items with a date format can be returned as the text representation of that date.

Chris-Stockbridge:
Fixes a bug where importing a sheet with a date colum that also
contained invalid dates would fail with a c++ exception and no other
info.
Adds detail to the error display when failing. Shows the user the text
that failed to import correctly along with the row and column numbers.
Rethrows the exception so that the process still fails to avoid
unanticipated behavior if someone doesn't look at the error messages.
Fixes #172

### DIFF
--- a/src/write_file.cpp
+++ b/src/write_file.cpp
@@ -228,7 +228,12 @@ SEXP buildMatrixMixed(CharacterVector v,
         }else{
           // dt_str = as<std::string>(m(ri,i));
           dt_str = m(ri,i);
-          datetmp[ri] = Rcpp::Date(atoi(dt_str.substr(5,2).c_str()), atoi(dt_str.substr(8,2).c_str()), atoi(dt_str.substr(0,4).c_str()) );
+          try{
+            datetmp[ri] = Rcpp::Date(atoi(dt_str.substr(5,2).c_str()), atoi(dt_str.substr(8,2).c_str()), atoi(dt_str.substr(0,4).c_str()) );
+          }catch(...) {
+            std::cerr << "Error reading date:\n" << dt_str << "\nrow: " << ri+1 << "\ncol: " << i+1 << "\n";
+            throw;
+          }
           //datetmp[ri] = Date(atoi(m(ri,i)) - originAdj);
           //datetmp[ri] = Date(as<std::string>(m(ri,i)));
         }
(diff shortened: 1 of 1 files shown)


src/write_file.cpp
```diff
@@ -228,7 +228,12 @@ SEXP buildMatrixMixed(CharacterVector v,
         }else{
           // dt_str = as<std::string>(m(ri,i));
           dt_str = m(ri,i);
-          datetmp[ri] = Rcpp::Date(atoi(dt_str.substr(5,2).c_str()), atoi(dt_str.substr(8,2).c_str()), atoi(dt_str.substr(0,4).c_str()) );
+          try{
+            datetmp[ri] = Rcpp::Date(atoi(dt_str.substr(5,2).c_str()), atoi(dt_str.substr(8,2).c_str()), atoi(dt_str.substr(0,4).c_str()) );
+          }catch(...) {
+            std::cerr << "Error reading date:\n" << dt_str << "\nrow: " << ri+1 << "\ncol: " << i+1 << "\n";
```
JanMarvin: Hi @ChrisSUMich `std::cerr` should be `Rcpp::Rcerr` see https://gallery.rcpp.org/articles/using-rcout/

Chris-Stockbridge: I made the change and pushed. I only have a little experience with github pull requests, but I think the same PR is now updated with the newest version. including changing to Rcpp::Rcerr

JanMarvin: Thanks, all fine. FYI: prior to this, R CMD check looked liked this:

```txt
  File 'openxlsx/libs/x64/openxlsx.dll':

  Found '_ZSt4cerr', possibly from 'std::cerr' (C++)

    Object: 'write_file.o'
```

PR #182, opened 2021-04-25
add final newline to Workbook show() method

### LINKED ISSUE #181: Workbook show() method does not end with a newline
When a Workbook object is shown in the console, the final newline is missing.

To reproduce:
```R
w <- openxlsx::loadWorkbook(system.file("extdata/readTest.xlsx", package="openxlsx"))
w
```
The final > prompt (cursor position) is somewhere in the middle of the screen, but should be on the left:

```
> w
A Workbook object.
[...]
Worksheet write order: 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12>
```

This minor bug affects R when run in a terminal or in Emacs, but RStudio seems to be more lenient about missing final newline, as shown with a simple `cat(pi)` test.

My proposed fix is to add `cat("\n")` at the end of the Workbook `show()` method - see pull request.

A missing newline is likely to affect all versions of R on all operating systems, but I have tested this in R 4.0.5 on Linux (Debian 10).

arni-magnusson:
Fixes #181

### DIFF
--- a/R/WorkbookClass.R
+++ b/R/WorkbookClass.R
@@ -3832,6 +3832,7 @@ Workbook$methods(
     }
 
     cat(unlist(showText))
+    cat("\n")
   }
 )
 


ycphs:
Thank you for your contribution.

PR #184, opened 2021-04-30
moves lookup tables to data-raw

jmbarbone:
This saves about 14,000 lines  in the `R\WorkbookClass.R` files and may provide an easier method to add more fonts within the csv file in `data-raw`.

* removes the data.frame creation from within the R\WorkbookClass.R file
* data.frame created and stored internally
* added a test for if the dimensions are changes
* makes a correction for a write.xlsx() test file extension


ycphs:
Currently there exists a merge conflict. But the concept to move the static data in a CSV file sounds good.

jmbarbone:
> Currently there exists a merge conflict. But the concept to move the static data in a CSV file sounds good.

Not exactly sure what it was picking up on but there didn't appear to be any _real_ conflicts with removing the data frame code.

Issue #187, opened 2021-05-04
write.xlsx doesn't accept argument tableName


bwelman:
Steps to reproduce:
write.xlsx(iris, file = "test.xlsx", asTable = TRUE, tableName = "mytable")

Version of openxlsx: 4.2.3

ycphs:
Could you please provide a file with the actual result and one with the expected?

bwelman:
library(openxlsx)
write.xlsx(iris, file = "irisdata.xlsx", asTable = TRUE, sheetName="data", tableName ="tableIris")

The result is an excel file with a correct sheetname but with tablename "Table3" and I expected a tablename "tableIris".
According to the documentation writeDataTable is used, and that function has an argument tableName.

PR #194, opened 2021-05-29
Writexlsx params

### LINKED ISSUE #192: New feature: buildWorkbook()
I've been toying with a bit of a rewrite of `write.xlsx()` which separates the building and saving process.  Ideally, a (named) list with `...` params could be passed to something like `buildWorkbook()` which would generate a `Workbook` object that can then be further modified before saving -- often something I personally find myself doing and could see a general utility.  `write.xlsx()` would then essentially be shorthand for ...

```r
wb <- buildWorkbook(data_list, ...)
saveWorkbook(wb, file)
```
... when finer edits would not be necessary.

The process of this would involve a better method of passing/matching parameters -- would would resolve issues like like #187 and wouldn't require additional edits when new params are added -- without changing the current use of `write.xlsx()`

jmbarbone:
Resolves #192 and #187.

Adds the `buildWorkbook()` function to extract the relevant processes from `write.xlsx()`.  The method in which I used should also allow for all applicable parameters passed through `...` into the other function calls.  

I have also made some formatting changes and small improvements in some of the code I was looking at. A large portion is simplifying the use of `grep()` and `grepl()` and adding some `assert_*()` functions which should make parameter checking easier and reduce some of the duplicate code.

It may be worth while, too, to update some of the tests that use `write.xlsx(x, file); wb <- read.xlsx(file)` as `buildWorkbook()` now accomplishes that.


R/readWorkbook.R
```diff
@@ -69,43 +69,47 @@
 #' }
```
JanMarvin, line 69: why did you modify this file?

R/wrappers.R
```diff
@@ -29,40 +29,40 @@
 #'   category = "something"
```
JanMarvin, line 29: same for this file. only whitespace changes?

R/writeData.R
```diff
@@ -14,6 +14,7 @@
 #' \code{c(startCol, startRow)}.
 #' @param colNames If \code{TRUE}, column names of x are written.
 #' @param rowNames If \code{TRUE}, data.frame row names of x are written.
+#' @param row.names Deprecated, please use \code{rowNames} instead
```
JanMarvin: I do not get this. This looks as if you added and deprecated it in the very same moment?

R/writeDataTable.R
```diff
@@ -12,6 +12,7 @@
 #' A vector of the form c(startCol, startRow)
 #' @param colNames If \code{TRUE}, column names of x are written.
 #' @param rowNames If \code{TRUE}, row names of x are written.
+#' @param row.names Deprecated, please use \code{rowNames} instead
```
JanMarvin: Same for this

JanMarvin (commented):
Coding looks nice and I like the wrappers for some of the functions, but I'm not gonna review a 3.5k diff of whitespace changes :smile:

R/writeData.R
jmbarbone: `row.names` had been originally passed from `write.xlsx()` and converted to `rowNames` before `writeData()` and `writeDataTable()`.  This was also done with `col.names` and `colNames` but I missed that because the latter wasn't used in testing.

R/writeDataTable.R
jmbarbone: As above.

R/readWorkbook.R
jmbarbone: Mostly from adding `assert_*()`s and while testing for any differences in changing up `write.xlsx()` as there were a lot of testing code that used the `write.xlsx(x, file); read.xlsx(file)` process

R/wrappers.R
jmbarbone: Some changes at  386.  Not entirely sure what all the white space changes are.  I'm using Rstudio `1.4.1103`, could there be some line return formatting causing this?

R/wrappers.R
```diff
@@ -383,91 +383,95 @@ addWorksheet <- function(
   od <- getOption("OutDec")
   options("OutDec" = ".")
   on.exit(expr = options("OutDec" = od), add = TRUE)
-  
+
```
jmbarbone, line 386: @JanMarvin some changes here

R/wrappers.R
JanMarvin: Rstudio has some formating options, removing trailing whitespace options, ensure empty line at the end of the file. For personal projects I usually stick with them, but all this unrelated noise make this pull request harder to read.
I'm not against the changes and from a quick glance they seem to improve the code, I just don't want to review all the lines of unrelated changes looking for possible issues in the new and changed code 😄

R/writeData.R
JanMarvin: I understand. I initially was wondering, why you'd deprecate functions from a stable API, but using similar arguments looks like an improvement. 👍

R/wrappers.R
jmbarbone: > Rstudio has some formating options, removing trailing whitespace options, ensure empty line at the end of the file. For personal projects I usually stick with them, but all this unrelated noise make this pull request harder to read.

Agreed.  I feel like this isn't going to be an isolated thing.  Temping now to make one big PR of just white-noise?

JanMarvin: First of, that's not my call to make (poking @ycphs) and ideally it would be a decision unrelated to an IDE configuration.
Personally, I would prefer to introduce such changes only when applying code changes. Iirc correctly, such whitespace related bulk commits make git bisecting and blaming harder, but _I didn't check that_.
Finally, I think that this should not be a blocker for this pull request, but I won't volunteer right now to review it. (And I know projects, where such pull request would be closed without any hesitation.)

ycphs: I totally agree with @JanMarvin. It is hard to read, but I will merge the PR. The next time it would be great if only the rows with actual changes is pushed.

PR #197, opened 2021-06-08
Minor lint issues

JanMarvin:


JanMarvin:
@ycphs this new lintr warning appeared: `Warning: local variable ‘new_ActiveSheet’ assigned but may not be used`. It's caused by

https://github.com/ycphs/openxlsx/blob/00fc559a8cb0db9a0cba23789812a8975963bbe9/R/wrappers.R#L2576

Presumably this line can be removed?

ycphs:
Yes you are right. I will remove the line.

PR #198, opened 2021-06-09
adds contributor

jmbarbone:
Adds me


ycphs (changes requested):
Malformed Authors@R field seems to be the issue

PR #208, opened 2021-06-25
Report failure when write.xlsx fails.

### LINKED ISSUE #190: write.xlsx does not error if it doesn't have write permissions
**Describe the bug**
If `write.xlsx` is asked to write out data to a file that the user doesn't have write permissions to, it fails silently.

**To Reproduce**
```
> fs::file_create("foo.xlsx", mode="a=rx")
> openxlsx::write.xlsx(x=cars, file="foo.xlsx", overwrite=TRUE)
```

**Expected behavior**
It should throw an error, like `write.csv` does:

```
> fs::file_create("foo.csv", mode="a=rx")
> write.csv(cars, file="foo.csv")
Error in file(file, ifelse(append, "a", "w")) : 
  cannot open the connection
In addition: Warning message:
In file(file, ifelse(append, "a", "w")) :
  cannot open file 'foo.csv': Permission denied
```

This was using the latest CRAN version: 4.2.3

jeff-m-sullivan:
Fixes #190 .

The pre-commit hooks were set up to require styler, so I ran styler on the files i changed, but committed the style pieces separately in case you want to strip them out.

Added a testthat test to show the bug and re-ran all tests to confirm that the fix doesn't break anything else.


tests/testthat/test-write-permissions.R
```diff
@@ -0,0 +1,14 @@
+
+context("error without write permissions")
+
+test_that("test failed write errors for saveWorkbook", {
+  tempFile <- fs::file_temp(ext = "xlsx")
+  fs::file_create(tempFile, mode = "a=rx")
```
jmbarbone: Could this be accomplished without important `fs`?

```r
tempFile <- temp_xlsx()
file.create(tempFile)
Sys.umask(file.mode(tempFile))
```

tests/testthat/test-write-permissions.R
```diff
@@ -0,0 +1,14 @@
...
+
+  wb <- createWorkbook()
+  addWorksheet(wb, "name")
+
+  expect_warning(write.xlsx(x = cars, file = tempFile, overwrite = TRUE))
```
jmbarbone: Maybe search for `"Permission denied"`?

R/wrappers.R
```diff
@@ -119,20 +119,15 @@ saveWorkbook <- function(wb, file, overwrite = FALSE, returnValue = FALSE) {
...
-    error = function(e) e, warning = function(w) w)
-
-
-
+  result <- file.copy(from = xlsx_file, to = file, overwrite = overwrite)
 
   ## delete temporary dir
   unlink(dirname(xlsx_file), force = TRUE, recursive = TRUE)
-  if(returnValue == FALSE){
+  if (returnValue == FALSE) {
```
jmbarbone, line 126: @ycphs 
Do you know if there any reason why we wouldn't want or be able to simply return the result of the `file.copy()`?  Does using `invisible(1)` satisfy anything?

ycphs: Initially there was no return value and  the feature request to have an optional made it necessary to have a default with no return value.

jmbarbone: The default value returned was `invisible(1)` which I don't see as useful as having the return value be the result of `file.copy()` instead -- especially since if `file.copy()` fails `saveWorkbook()` essentially fails.  So I guess I'm asking: should `invisible(1)` _ever_ be returned or should the result of  just be `file.copy()` (invisibly) returned?

ycphs: yes that would make more sense and it helps to actually and leads to the initial error.

tests/testthat/test-write-permissions.R
jeff-m-sullivan: Unfortunately, `file.create` doesn't allow for setting permissions, so I chose to use `fs::file_create` instead of using `file.create` and then having platform-specific commands to set the read-only status.

tests/testthat/test-write-permissions.R
jeff-m-sullivan: Great idea. I'll add that in.

jeff-m-sullivan: Implemented in ed1f0d8dd39d6e536ff8902b071d6c3b78cefe78

tests/testthat/test-write-permissions.R
ycphs: I am not sure if it is necessary to have a new dependency for a function used in the test.

jeff-m-sullivan: Should I remove the dependency and just have the GitHub workflows install the fs package explicitly for those which run the tests?

jeff-m-sullivan: I think I found another way. I'll update the branch later this evening.

jeff-m-sullivan: Fixed in ec229fea1de5033c9cf25baebe0a019a07717998

Issue #209, opened 2021-06-29
read.xlsx() can no longer parse .xlsm files


dseynaev:
we were using openxlsx `4.2.3` before and successfully parsing `.xlsm` files with `read.xlsx`

in `4.2.4` we see:
```
openxlsx can only read .xlsx files
```

my guess is this was introduced here:
https://github.com/ycphs/openxlsx/commit/86070beda7d12defa1c0fba622c2f58312d1256c#diff-43a394a441eb434c8b24866cf52453aa6dbca568a67a60e98d08ebaab8124603L119-L121

is it intended for `read.xlsx` to not support `.xlsm`? or was the regex pattern improperly generalized?

JanMarvin:
Hi @dseynaev , the change was most likely unintended. Could you create a pull request to modify this regex?

ycphs:
I think we can close the issue

jormercury:
Thanks for fixing this, I had to instruct several people to downgrade to 4.2.3. Is it possible to release a stable version with this fix soon? It seems to me like a simple fix, but with large consequences for anyone trying to read in xlsm files using openxlsx.

Issue #205, opened 2021-06-22
Cannot read .xlsm files


david-f1976:
`read.xlsx()` has been updated so that an error is produced if the file extension of the file being read is not .xlsx. I have been reading .xlsm files without issue in prior releases.

This is on 4.2.4 of openxlsx; it was working on 4.2.3.

ycphs:
If you install the development version, the bug should be fixed.

```
install.packages(c("Rcpp", "devtools"), dependencies = TRUE)
require(devtools)
install_github("ycphs/openxlsx")
```

david-f1976:
Working here now thanks

PR #210, opened 2021-06-29
fix: extend regex in read.xlsx to allow .xlsm files; solves #209

dseynaev:
solves https://github.com/ycphs/openxlsx/issues/209


JanMarvin:
also #205

JanMarvin (approved):

Issue #206, opened 2021-06-24
write.xlsx "_xb025_" result error


ghost:
**Describe the bug**

write.xlsx xb025 result error

**To Reproduce**

```
test_char <- "_topic_ext_xb025_tm"
df <- data.frame(col = test_char)
openxlsx::write.xlsx(df, "~/Downloads/test_char.xlsx")
```

**Expected behavior**
It should be save "_topic_ext_xb025_tm"

**Screenshots**
It should be like this:
<img width="259" alt="image" src="https://user-images.githubusercontent.com/30428489/123189475-72d85b80-d4d0-11eb-80a8-96ce0af55c83.png">

Actually result: Error

<img width="207" alt="image" src="https://user-images.githubusercontent.com/30428489/123189496-7bc92d00-d4d0-11eb-8869-a074e3f21176.png">

**Desktop (please complete the following information):**
 - OS: macOS Big Sur 11.4
 - openxlsx version: Documentation for package ‘openxlsx’ version 4.2.4

ycphs:
it seems that the issue only appears with Microsoft office. libreoffice doesn't have this behavior.

Maybe it is a bit more complicated.

ghost:
> it seems that the issue only appears with Microsoft office. libreoffice doesn't have this behavior.
> 
> Maybe it is a bit more complicated.

Could it be Microsoft's problem?

ycphs:
At least it looks like that.

ycphs:
As far as I checked it is an issue just with Micrcosoft.

Issue #152, opened 2021-03-25
Writing certain text (unicode I suppose) corrupts the file


aavanesy:
**Describe the bug**
See code. For some reasong if a text contains '\037' it will write but not open the file.
I haven't tested other strings but this is I suppose some unicode issue.

**To Reproduce**
```wb <- openxlsx::createWorkbook()
text = 'abcd\037'
openxlsx::addWorksheet(wb, 'Data')
writeData(wb, 'Data', text, startRow = 1)
saveWorkbook(wb,'file3.xlsx', overwrite = T)
```

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows 8.1
 - R 4.02
 - Openxlsx 4.2.3

**Additional context**

bryburgess:
Same issue on MacOS 11.2.3, R 4.02, Openxlsx 4.2.3. 

I tried writing a list of tibbles and keep encountering the string error. All the tibbles have characters in the first col, and the rest are numeric with a mix of NAs/values. Excel won't open the file without force-fixing it, which just ends up stripping the column names, first character column, and the explicit NAs from the data. 

**Screenshot**:
<img width="711" alt="Screen Shot 2021-06-04 at 2 26 35 AM" src="https://user-images.githubusercontent.com/85198735/120755769-69b12b80-c4dc-11eb-8ccf-c9025b358998.png">

**Excel error message:**

> 
> > <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
[...]

PR #214, opened 2021-06-30
unicode escape for shared strings

ycphs:
add and remove escape (_x005F) for format _xhhhh_

### DIFF
--- a/R/WorkbookClass.R
+++ b/R/WorkbookClass.R
@@ -1107,6 +1107,8 @@ Workbook$methods(
   updateSharedStrings = function(uNewStr) {
     ## Function will return named list of references to new strings
     uStr <- uNewStr[which(!uNewStr %in% sharedStrings)]
+   
+    
     uCount <- attr(sharedStrings, "uniqueCount")
     sharedStrings <<- append(sharedStrings, uStr)
 
--- a/R/helperFunctions.R
+++ b/R/helperFunctions.R
@@ -386,6 +386,37 @@ replaceIllegalCharacters <- function(v) {
 }
 
 
+
+add_escape_unicode <- function(v) {
+ 
+  strings_unicode <-
+    c(stri_extract_all_regex(v, "_x+[:xdigit:]{4}+_", simplify = T))
+  
+  strings_unicode_clean <- paste0("_x005F", strings_unicode)
+  
+  v <- stri_replace_all_fixed(v, strings_unicode, strings_unicode_clean)
+  
+  
+  
+  
+  return(v)
+}
+
+
+
+remove_escape_unicode <- function(v) {
+  strings_unicode <-
+    c(stri_extract_all_regex(newstring, "_x005F_x[:xdigit:]{4}_", simplify = T))
+  
+  strings_unicode_clean <-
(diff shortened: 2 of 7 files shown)


ycphs:
fix for #206

JanMarvin:
Hi @ycphs , did you figure out, what's causing this behavior in the first place? I assume it's one of of the many loops we have to jump through, to provide UTF-8?

ycphs:
Hi @JanMarvin ,

It has nothing to do with the utf8 conversion. It's an "Excel" feature converting all strings in the format `_xhhhh_` to the unicode character, if not escaped with `_x005f`.

I just found some hints at SO like [thread](https://stackoverflow.com/questions/47515917/spreadsheetdocument-openxml-decoding-the-encoded-text-in-excel-file).

I tried to check the behaviour on libre office, but I didn't had the expected results. Any input is welcomed!

JanMarvin:
Got it, thanks for the explanation!

ycphs:
It seems Microsoft Excel has the issue, but with libreoffice calc the encoding issue doesn't exist

ycphs:
I have to find a way to get the standard app for .xlsx files. Based on that an additional option could be set. 

Basically Microsoft Excel with the strange escape and others like libre office without the escape. This impacts the read and the write functions.

JanMarvin:
iirc there are additional sheet names in app.xml. maybe these are used too? and even though it's possible, is it really worth the hassle? if the user can circumvent it, shouldn't we just document it and leave it be? otherwise maybe someone can provide an xlsx file that behaves identical in Excel and Libreoffice?

JanMarvin:
Something like this should do the trick. The regex is from [Microsoft](https://github.com/dotnet/wcf/blob/1a291ea424991a184ffafd2ba20928e0e339fb96/src/dotnet-svcutil/lib/src/FrameworkFork/Microsoft.Xml/Xml/XmlConvert.cs#L212) and apparently it checks for strings that look like unicode chars utf-16 and utf-32. The entire idea seems to be, that it guards against unwanted unicode conversion.
```R

string <- "_topic_ext_xb025_tm" #original issue
string <- "_topic_ext_Xb0288886_tm" # test with 8 chars after _x
expr <- "(?<=_)[Xx]([0-9a-fA-F]{4}|[0-9a-fA-F]{8})_"

# convert to matrix n x 2
get <- matrix(unlist(stringi::stri_match_all_regex(pattern = expr, str = string)), ncol=2)

# add unicode _ as prefix ("x005F")
get[,2] <- paste0("x005F_x", get[,2], "_")

# replace in string
stringi::stri_replace_all_fixed(string, pattern = get[,1], replacement = get[,2], vectorize_all = F)
```

Didn't test the impact in Excel or LibreOffice, but with .NETs `XmlConvert.EncodeName` [here](https://dotnetfiddle.net/eEFrCs). The actual EncodeName code is not only the regex, but I didn't try to understand the EncodeName code :smile: (apparently it encodes other cases as well: e.g. if the name begins with a number, the first digit is encoded into its unicode representation 1 becomes `_x0031` and other signs notably `>` is converted too; see also `Unicode::as.u_char(utf8ToInt("_"))`)

ycphs:
#152

PR #216, opened 2021-07-02
resolves #215

### LINKED ISSUE #215: openxlsx_setOp() fails with named list
This was meant to implemented but there is a missing `return()` for the `mapply()`.  Will be sending a PR for a fix with tests.

https://github.com/ycphs/openxlsx/blob/master/R/openxlsx.R#L134-L146

``` r
library(openxlsx)

x <- list(withFilter = TRUE, keepNA = TRUE)
openxlsx_setOp(x)
#> Error in as.list(value): argument "value" is missing, with no default
```

<sup>Created on 2021-07-02 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

jmbarbone:
* adds correct return() for the list functions
* invisible() used to match the outputs for non-list
* cleans up the test-options.R

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # openxlsx (development version)
 
+## Fixes
+
+* `openxlsx_setOp()` now works with named list ([#215](https://github.com/ycphs/openxlsx/issues/215))
+
 # openxlsx 4.2.4
 
 ## Fixes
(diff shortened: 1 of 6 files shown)


ycphs:
I had to revert the commit, because the checks failed. Could you please create a new PR?

Issue #86, opened 2020-07-31
Delete/Overwrite named regions functionality


HelenwoodsJNCC:
Hi, 

The deleteNamedRegions function was announced for version 4.1.0 but doesn't appear to have been included. Could you please advise on the status of this functionality? 

Thanks

ycphs:
It is in the news, but I couldn't find any function to cover the functionality. Maybe it was a planned feature, but I can't find it in the former or current repo.

I am happy, if you could provide a pull request to add the functionality to the package.

PR #223, opened 2021-07-08
#86 add Delete/Overwrite named regions functionality

bthieurmel:
Add : 

- new parameter ``overwrite = FALSE`` in ``createNameRegions``
- and new function ``deleteNameRegions``


bthieurmel:
Really not so must change, only few lines (not 600 as git say...!)

image

image

+ doc

ycphs:
I thought of implementing also the printing region. And also a check if the named region overwrites it.

ycphs:
Feature request for editing the print area 

#235

ycphs:
> Really not so must change, only few lines (not 600 as git say...!)
> 
> image
> 
> image
> 
> * doc

Could you please create a new PR without the white space changes.

bthieurmel:
Yes. PR without space here #251

Issue #228, opened 2021-07-14
Version 4.2.4 cause text rotation formating lost on saving workbook.


GordonAn:
**Version 4.2.4 cause  text rotation formating lost on saving workbook**

After loading a workbook with text rotation 90 degree,   Direct saving the workbook without any change. The resulted Excel file, when opened in Excel, the text become horizontal and column width also changed.  In versino 4.2.3, the text rotation formating is kept without any problem.   I am using R on windows, and with Excel 360.

** The following 3 lines of codes will reproduce the issue**

library("openxlsx")  
packageVersion("openxlsx")  #Needs version4.2.4 to reproduce the issue
 
Wb<-loadWorkbook("InputTestFile_V424NotkeepTextRotation.xlsx")
saveWorkbook(Wb, file="Output_V424NotkeepTextRotation.xlsx", overwrite = TRUE, returnValue = TRUE)
#Verision 4.2.3 works fine
#Version 4.2.4 lost text rotation format and column Width
 

**Expected behavior**
The text rotation formatting be kept.

**Screenshots**
If applicable, add screenshots to help ex
![lost text rotation on saving](https://user-images.githubusercontent.com/30178216/125599720-4d97f415-b49c-4fc2-a637-0586bae1d319.jpg)
plain your problem.

**Desktop (please complete the following information):**
 - OS:  Windows
  
The attached is the input excel and the output 
[InputTestFile_V424NotkeepTextRotation.xlsx](https://github.com/ycphs/openxlsx/files/6815037/InputTestFile_V424NotkeepTextRotation.xlsx)
[Output_V424NotkeepTextRotation.xlsx](https://github.com/ycphs/openxlsx/files/6815039/Output_V424NotkeepTextRotation.xlsx)

gigsterdev9:
This happened to me too. Existing formatting is altered on an Excel file once a workbook is opened and then saved on to. Rolling back to 4.2.3 appear to have addressed the issue for me.

s-fleck:
I'm having the same issues, for me even very basic formatting (bold, centred, etc...) gets lost

JanMarvin:
A potential fix is in #237 could any of you test this please?

JanMarvin:
Should be fixed in https://github.com/ycphs/openxlsx/commit/0fc1c08ceebfa6a9fd50cd54c4bb55459b2900a5

Issue #207, opened 2021-06-24
loadWorkbook removing formatting


mdorchuckuva:
When using loadWorkbook, formatting that has been applied by a user in excel directly is being stripped.  

With the attached excel file in my working directory, running the following code opens a file with cell C4 in black, unbolded text instead of the red, bold text as originally in the file.  The formatting remains stripped if I save the wb object back out and then open from file explorer. 

```
library(openxlsx)
wb = loadWorkbook("testopenxlsx.xlsx")
openXL(WB)
```

System information
Windows 10 Enterprise 10.0.19042
Have tested on R 4.0.3 and R 4.1.0, in both cases the issue persists
Openxlsx version 4.2.4
Office 365 version 2102
[testopenxlsx.xlsx](https://github.com/ycphs/openxlsx/files/6711990/testopenxlsx.xlsx)

mdorchuckuva:
I found that the issue no longer occurs if I back up my openxlsx version to 4.0.0, in case that helps narrow it down.

ycphs:
In your example is you open a different workbook than the one you loaded

```
library(openxlsx) 
wb = loadWorkbook("testopenxlsx.xlsx")
openXL(WB)
```

Please try the following:


```
library(openxlsx) 
wb = loadWorkbook("testopenxlsx.xlsx")
openXL(wb)
```

mdorchuckuva:
Sorry the wb got capitalized when I pasted over for some reason.  I was running the same workbook.  see screenshot below and thanks so much for taking a look!

Top workbook is what opens from OpenXL, bottom workbook is the workbook that was originally saved.  

![openxlsx_screenshot](https://user-images.githubusercontent.com/27371304/123335751-e8fdb200-d512-11eb-8faf-6514e77760b1.png)

JanMarvin:
I've seen similar issues before. What I have found is that the style order is changed by `openxlsx`. We always add our default style to `styles.xml`, which is picked here and all the other styles are re-sorted and do no longer match with the style ids in the worksheet. Iirc the styles file has no style id field and they are picked in the order of appearance. Kinda the same issue as #81.

Seviks:
I can confirm this issue on 4.2.4. When I `loadWorkbook`, do my changes (or not) an then `saveWorkbook`, the formatting is gone. 

Formats persisted in 4.2.3. After roll back to 4.2.3 the issue is gone.

Tested on:
openxlsx 4.2.4
Windows 10
R 4.1.0
[...]

PR #237, opened 2021-07-30
[WIP] Fixes a regression in 4.2.4 and restores loading <foo/>. closes # 207

JanMarvin:
Fixes reading `<b/>` tag in font. Previously only `<b val="true"/>` was imported, but `<b/>` is valid too. I assume this was caused by my last attempt to fix this function. Needs some unittests and maybe someone should check if this actually fixes the error *and* what precisely `getChildlessNode()` should do.

A remaining question is: are multiple values supported e.g. should these two be detected and returned: `openxlsx:::getChildlessNode("<b foo=\"1\"/><b foo=\"1\"/><b/>", "b")`? The current implementation works like this for cases unlike `<foo/>`. With `<foo/>` only a a single string is returned. For `<fonts>` this shouldn't matter as all `<font>` nodes are separated before `getChildlessNode()` is called.

Ultimately this _needs_ testing.

This is related to #207 and ~~maybe~~ #228 ~~(but I didn't test this)~~.

Some lines to test, should be converted to a test:
```R
# tests, 1 indicates a result is returned
openxlsx:::getChildlessNode("<b/>", "b") # 1
openxlsx:::getChildlessNode("<b foo=\"1\"/><b/>", "b") # 1
openxlsx:::getChildlessNode("<b>foo</b>", "b") # 0
openxlsx:::getChildlessNode("<ba/>", "b") # 0
openxlsx:::getChildlessNode("<ba foo=\"1\"/>", "b") #0
```

Note to self:
```R
# Most likely something for another issue. File is from #232 
wb <- loadWorkbook("~/font_test.xlsx")
saveWorkbook(wb, "/tmp/test.xlsx", TRUE)

# This produces a broken file
wb2 <- loadWorkbook("/tmp/test.xlsx")
saveWorkbook(wb2, "/tmp/test2.xlsx", overwrite = TRUE)
```

### DIFF
--- a/R/WorkbookClass.R
+++ b/R/WorkbookClass.R
@@ -4061,7 +4061,7 @@ Workbook$methods(
             }
 
             flags <-
-              c("bold", "italic", "underline") %in% names(thisFont)
+              c("bold", "italic", "underline", "strikeout") %in% names(thisFont)
             if (any(flags)) {
               style$fontDecoration <- NULL
               if (flags[[1]]) {
@@ -4077,6 +4077,11 @@ Workbook$methods(
               if (flags[[3]]) {
                 style$fontDecoration <-
                   append(style$fontDecoration, "UNDERLINE")
+              }
+
+              if (flags[[4]]) {
+                style$fontDecoration <-
+                  append(style$fontDecoration, "STRIKEOUT")
               }
             }
           }
--- a/R/WorkbookClass.R
+++ b/R/WorkbookClass.R
@@ -1640,9 +1640,9 @@ Workbook$methods(
   getBaseFont = function() {
     baseFont <- styles$fonts[[1]]
 
-    sz <- getAttrs(baseFont, "<sz ")
-    colour <- getAttrs(baseFont, "<color ")
-    name <- getAttrs(baseFont, "<name ")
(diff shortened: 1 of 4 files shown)


JanMarvin:
I assume this fixes the regression and I assume that every load and save action from `openxlsx` might be impacted. The issue was most likely caused by this commit https://github.com/ycphs/openxlsx/commit/a06b6bc796f5e6e436cb6e33b2a1f8b5d3685cfa by yours truly. Iirc I attempted to fix `definedNames` and did not fully understand the impact of what I was doing. A side effect of my commit broke imports of `<foo/>` and with `getChildlessNode()` and `getAttrs()`.
This causes the font selection to be completely broken since tags like `<b/>` were no longer imported correctly and other style attributes are affected too. My assumption is that styles are lost and or randomly reapplied. 

This actually might have caused me a bunch of headaches and many sleepless nights that led to #134 (me wrongfully assuming that `openxlsx` xlsx files look kinda broken).

@ycphs Once this PR is tested, we might consider releasing a hotfix to CRAN.

ycphs:
@JanMarvin 
Yes a CRAN release is a good idea.

PR #238, opened 2021-08-02
add stale.yaml

JanMarvin:
As discussed, adds github action stale to close issues after a year


JanMarvin:
merging because it's only affecting GitHub

PR #250, opened 2021-08-26
Fixed list element names

### LINKED ISSUE #248: v4.2.4 breaks number formatting with borders
**Describe the bug**
`write.xlsx(x, file)`-function crashes when `x` contains a column with class `percentage` or `accounting` and an optional parameter `borders = "columns"` is used.

Error message:
```r
Error in if (this.sty$numFmt$numFmtId == 9999) { : 
  argument is of length zero
```

**Expected behavior**
An Excel file output with percent/accounting-formatted column with borders.

**Troubleshooting**
It looks like commit e33ab30e4554e178441e97378d3dce9febdfec8d (v4.2.4) breaks the functionality in `R/borderFunctions.R` due to a possible typo in `colStyle$numFmt`. In the list, there is `numFmtIs = ...` that should be – at least as far as I understand – `numFmtId = ...`. The same applies for both accounting and percentage styles, whereas the others seem to be fine.

epanormaalijakauma:
Typos in accounting and percentage formats in R/borderFunctions.R

Fixes #248

### DIFF
--- a/R/borderFunctions.R
+++ b/R/borderFunctions.R
@@ -10,11 +10,11 @@ genBaseColStyle <- function(cc) {
   } else if ("currency" %in% cc) {
     colStyle$numFmt <- list(numFmtId = "164", "formatCode" = "&quot;$&quot;#,##0.00")
   } else if ("accounting" %in% cc) {
-    colStyle$numFmt <- list(numFmtIs = "44")
+    colStyle$numFmt <- list(numFmtId = "44")
   } else if ("hyperlink" %in% cc) {
     colStyle$fontColour <- list(theme = "10")
   } else if ("percentage" %in% cc) {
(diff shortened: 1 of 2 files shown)


JanMarvin:
Thanks, looks good to me!

JanMarvin:
PS: You could add your minimal example as a test, so that we're not falling into this pit in the future.

ycphs:
Thank you

epanormaalijakauma:
@JanMarvin, I haven't written any automated tests in R packages yet, so I'll leave it to you or the others in the community. Feel free to use my example in #248 as a part of testing though.

JanMarvin:
Thanks for the reply. I've added a test to the pull request. Once it's greenlighted it can be merged.

Issue #256, opened 2021-09-10
Add Accounting Underline Style to TextDecoration


rfineman:
In addition to single and double underlining a cell, it would be great if we could have the options: "Single Accounting" and "Double Accounting."

<img width="176" alt="underline" src="https://user-images.githubusercontent.com/33735858/132872567-29cfb194-a146-4206-bf53-688a4b6e290a.PNG">

PR #258, opened 2021-09-10
Accounting underline

rfineman:
Fixes Issue #256
add single accounting and double accounting underline option to textDecoration parameter in createStyle.
Redoing pull to deal with odd diff error.


JanMarvin:
Thanks, now only the man page needs an update. Otherwise looking good (haven't checked it yet in Excel/LO, but it should not break anything).

rfineman:
Ok updated createStyle man

ycphs:
Great @rfineman , thank you.

Issue #262, opened 2021-09-27
Allow parallelization of getSheetNames


sgrote:
**Is your feature request related to a problem? Please describe.**
It does not seem to be possible to run `openxlsx::getSheetNames` in parallel. Here's my code:

```r
getSheetNames <- function(excel_files, parallel = FALSE, BPPARAM = BiocParallel::bpparam()) {
  if (parallel) {
    sheet_list <- BiocParallel::bplapply(excel_files,
      FUN = function(excel_file) {
        if (!file.exists(excel_file)) {
          return(NULL)
        }
        openxlsx::getSheetNames(excel_file)
      },
      BPPARAM = BPPARAM
    )
  } else {
    sheet_list <- lapply(excel_files,
      FUN = function(excel_file) {
        if (!file.exists(excel_file)) {
          warning("File '", excel_file, "' could not be found.")
          return(NULL)
        }
        openxlsx::getSheetNames(excel_file)
      }
    )
  }
  names(sheet_list) <- excel_files
  sheet_list
}
```
 Now, `getSheetNames(excel_files, parallel = TRUE)` sometimes throws an error

```r
 Error: BiocParallel errors
  element index: 1
  first error: cannot open file '/tmp/Rtmpxttndw/_excelXMLRead/[Content_Types].xml': No such file or directory 
```
or
```r
Error: BiocParallel errors
  element index: 1
  first error: cannot open the connection 
```

**Describe the solution you'd like**
My guess would be that each of the parallel tasks writes the same temporary file.  
Maybe using a directory or a file based on `tempfile()` would be a solution.

**Describe alternatives you've considered**
Parallelization does work with `readxl::excel_sheets`, but that is much slower than `openxlsx::getSheetNames`.
Another alternative is to not run it in parallel, but with a growing number of excel files this will be more time consuming.

jmbarbone:
I'm not familiar with BiocParallel but I don't have an issue trying to read with the `parallel` package and the current dev version of `openxlsx`

``` r
library(openxlsx)
library(parallel)
packageVersion("openxlsx")
#> [1] '4.2.4.9000'

files <- sapply(1:10, function(x) tempfile(fileext = ".xlsx"))
invisible(sapply(files, function(x) {
  write.xlsx(list(a = data.frame(1)), x)
}))

cl <- makeCluster(2)
res <- parSapply(cl, files, getSheetNames)
stopCluster(cl)

names(res) <- basename(names(res))
res
[...]

PR #263, opened 2021-09-28
changes in temp directory related to #262

jmbarbone:
Although I couldn't replicate the issues with #262 there are some instances in which creating a temporary directory can be simplified.  It doesn't seem like the directory names have to be specifics as tests passed locally.

@sgrote can you try your code with these changes and see if you're still encountering issues?


sgrote:
I installed this version and can confirm that my issue is solved. Thanks a lot! :)

JanMarvin:
Thanks!

PR #270, opened 2021-10-12
Update GitHub Actions

JanMarvin:
This updates most of the Github Actions. Most of it is copy and paste from their examples.

For R CMD check I've removed a few older releases. Personally I'm not a huge fan of testing everything everywhere. CI is a blessing, but can become a curse if things start failing for unknown reasons.

@jmbarbone happy if you could have a look too.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -20,13 +20,13 @@ jobs:
       fail-fast: false
       matrix:
         config:
-          - {os: windows-latest, r: 'oldrel'}
+          - {os: windows-latest, r: 'oldrel-1'}
           - {os: windows-latest, r: 'release'}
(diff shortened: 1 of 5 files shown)


ycphs (approved):

jmbarbone:
@JanMarvin only thing I'd recommend is considering testing down to R 3.3.0 as it's in the depends.

https://github.com/ycphs/openxlsx/blob/24781e4ed483723c7445574610a37e4f9df84ecb/DESCRIPTION#L39-L40

Looks like the old tests went down to 3.6 which could be good to keep and update the depends to this requirement.

JanMarvin:
Thanks, valid point. Though in this case I'd be in favor of dropping the depends on a specific R version if that's possible.
I need the CI to point me to code that breaks things and not to broken test systems :smile:

PR #274, opened 2021-10-16
col2int implementation (similar to int2col) . closes #273

### LINKED ISSUE #273: Function to convert column letters in Excel into integers
I use openxlsx and show it to colleagues for many years and each time we discuss with someone that it would be great to specify columns in read.xlsx() as letters, not numbers. This fact is the only reason I recently started to show people also the package readxl, where you can easily specify a range like 'A3:G11' for import (but no possibility to omit certain rows or columns from the range prevents me from showing readxl only). Of course, R1C1 style in Excel is a workaround to quickly find out the column numbers, but it is not convenient at all.

I've just figured out that there is the function int2col() in openxlsx and some function like col2int() with kind of inverted behavior would be very convenient to use for that (could not find anything like this in the package). I would expect supplying this function with a single string like "B:F, K, O, X:AD".

Would it be possible to add such a function to the package? Thanks a lot!

JanMarvin:
match `int2col()`

* [x] Needs a unittest


JanMarvin:
I'll add the unittest sometime next week

R/wrappers.R
```diff
@@ -207,6 +207,25 @@ int2col <- function(x) {
...
+#' @param x A character vector
+#' @export
+#' @examples
+#' col2int(LETTERS)
+col2int <- function(x) {
+  
+  if (!is.character(x)) {
+    stop("x must be character")
+    
+    if (any(is.na(x)))
```
JanMarvin: this is never reached

ycphs: true

ycphs (approved):
for me that's fine

PR #276, opened 2021-10-17
275 create comment

### LINKED ISSUE #275: `createComment(width)` fails with integers
Reported in [awalker89/openxlsx/issues/538](https://github.com/awalker89/openxlsx/issues/538) by @yb2125

Class checking for comment is too strict/not correct.

``` r
library(openxlsx)
createComment("hi", width = 2)
createComment("hi", width = 2L)
#> Error in createComment("hi", width = 2L): width argument must be a numeric vector
```

<sup>Created on 2021-10-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

jmbarbone:
Closes #275

* Makes changes to parameter checking to allow for `integers` (as documented)
* Updates docs to be clear on how `height` and `width` should be entered
* Adds checking for single values and removes the lines that grab the first values of height, width without error/warning
* Simplifies other checks for comments
* Adds unit testing

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -16,7 +16,7 @@
 * `makeHyperlinkString()` does no longer require a sheet argument ([#57](https://github.com/ychps/openxlsx/issues/57), [#58](https://github.com/ychps/openxlsx/issues/58))
 * improvements in how `openxlsx` creates temporary directories (see [#262](https://github.com/ychps/openxlsx/issues/262))
 * `writeData()` calls `force(x)` to evaluate the object before options are set ([#264](https://github.com/ycphs/openxlsx/issues/264))
-* `createComment()` how correctly handles `integers` in `width` and `height` ([#275](https://github.com/ycphs/openxlsx/issues/275))
+* `createComment()` now correctly handles `integers` in `width` and `height` ([#275](https://github.com/ycphs/openxlsx/issues/275))
 
 # openxlsx 4.2.4
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -16,6 +16,7 @@
 * `makeHyperlinkString()` does no longer require a sheet argument ([#57](https://github.com/ychps/openxlsx/issues/57), [#58](https://github.com/ychps/openxlsx/issues/58))
 * improvements in how `openxlsx` creates temporary directories (see [#262](https://github.com/ychps/openxlsx/issues/262))
 * `writeData()` calls `force(x)` to evaluate the object before options are set ([#264](https://github.com/ycphs/openxlsx/issues/264))
+* `createComment()` how correctly handles `integers` in `width` and `height` ([#275](https://github.com/ycphs/openxlsx/issues/275))
 
 # openxlsx 4.2.4
 
--- a/R/asserts.R
+++ b/R/asserts.R
@@ -67,7 +67,7 @@ assert_numeric1 <- function(x, scalar = FALSE) {
   ok <- is.numeric(x) & length(x) == 1L
   
   if (scalar) {
-    ok <- ok && length(x) == 1L
+    ok <- ok && nchar(x) == 1L
     msg <- paste0(msg, "single number")
   } else {
     msg <- paste0(msg, "numeric vector of length 1L")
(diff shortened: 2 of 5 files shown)


NEWS.md
```diff
@@ -16,6 +16,7 @@
 * `makeHyperlinkString()` does no longer require a sheet argument ([#57](https://github.com/ychps/openxlsx/issues/57), [#58](https://github.com/ychps/openxlsx/issues/58))
 * improvements in how `openxlsx` creates temporary directories (see [#262](https://github.com/ychps/openxlsx/issues/262))
 * `writeData()` calls `force(x)` to evaluate the object before options are set ([#264](https://github.com/ycphs/openxlsx/issues/264))
+* `createComment()` how correctly handles `integers` in `width` and `height` ([#275](https://github.com/ycphs/openxlsx/issues/275))
```
JanMarvin: typo should be "now"

R/asserts.R
```diff
@@ -62,6 +62,22 @@ assert_unique <- function(x, case_sensitive = TRUE) {
   }
 }
 
+assert_numeric1 <- function(x, scalar = FALSE) {
```
JanMarvin, line 65: Even though the scalar option is not called in the code (do we need it), can we use something like this [`is.scalar()`](https://stackoverflow.com/a/38088874/12340029)

tests/testthat/test-CommentClass.R
```diff
@@ -0,0 +1,18 @@
...
+  expect_error(createComment("hi", height = 1:2), "height")
+  
+  expect_error(createComment("hi", visible = NULL))
+  expect_error(createComment("hi", visible = c(TRUE, FALSE)), "visible")
+  
+  expect_error(createComment("hi", author = 1))
+  expect_error(createComment("hi", author = c("a", "a")), "author")
+  
+  expect_s4_class(createComment("hello"), "Comment")
+})
```
JanMarvin, line 18: Could you add a pending empty line at the end of the file? Otherwise `lintr` will complain and I have sleepless nights

JanMarvin (approved):
As usual, only minor nitpicking - otherwise looking good!

tests/testthat/test-CommentClass.R
JanMarvin: Oh maybe there is one already and github simply hides it

R/asserts.R
jmbarbone: I added that in to match the `assert_character1()` arguments.  I'd be fine with using something like that suggestion, especially if we have other parameters that we want to consistently check.  The suggestion would also work a bit better with `""`.

``` r
is.atomic("") && length("") == 1L
#> [1] TRUE
nchar("") == 1L
#> [1] FALSE
```

<sup>Created on 2021-10-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

tests/testthat/test-CommentClass.R
jmbarbone: Yup, I think this one is hidden.  I think maybe Github will warn when there isn't a new line?  At least the GitHub Desktop app I uses does that.

R/asserts.R
jmbarbone: Ah, actually I'm not quite right.  The scalar option for the character check was specific for this:

https://github.com/ycphs/openxlsx/blob/a4f9b83900cba9a6044005a5ce516f47f22adf13/R/readWorkbook.R#L132

... where the documentation demanded a **single** character.

``` r
openxlsx:::assert_character1("aa")
try(openxlsx:::assert_character1("aa", scalar = TRUE))
#> Error : aa must be a character vector of length 1L
openxlsx:::assert_character1("a", scalar = TRUE)

openxlsx:::assert_numeric1(11)
try(openxlsx:::assert_numeric1(11, scalar = TRUE))
#> Error : 11 must be a single number
openxlsx:::assert_numeric1(1, scalar = TRUE)
```

<sup>Created on 2021-10-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Perhaps "scalar" may not be the best word?  Changing it shouldn't be an issue since they're just internal functions.

JanMarvin: well, we have all the time in the world to come up with a better name 🙂

JanMarvin:
Thanks, merged!

PR #282, opened 2021-10-28
fix visibility issue. fixes #279

### LINKED ISSUE #279: Hidden sheets unhide when loading and saving
- Tab formatting: hidden tabs reappear upon loading and saving the workbook.

_Originally posted by @mendy-friedman in https://github.com/ycphs/openxlsx/issues/207#issuecomment-952166666_

JanMarvin:
This PR fixes an issue where hidden sheets would unhide after loading and saving a workbook with `openxlsx`. The issue was caused by the following line which returned an incorrect vector.

```R 
unlist(strsplit(sheets, split = "<sheet ")[-1])
```

The bug is located here:

https://github.com/ycphs/openxlsx/blob/f890014ee4fc31bce2bf8cb5daf8229ba9bec350/R/loadWorkbook.R#L177-L180

Since `is_visble` was most likely of some other length than `sheetId`, it is "corrected" to return "visible" for all sheets. This is assigned to the workbook object here: 

https://github.com/ycphs/openxlsx/blob/f890014ee4fc31bce2bf8cb5daf8229ba9bec350/R/loadWorkbook.R#L209

The faulty line might have been correct once (though it looks kinda strange). Anyways, since our node selection has improved, hopefully the `strsplit` isn't required anymore.

We might want to add a test for this.


JanMarvin:
@mendy-friedman you might wanna check if this solves your visibility issue.

mendy-friedman:
Thank you for addressing this so quickly - I'm traveling this week but will check later for my production script.  I did run a quick test on a simpler file and it seems to be working.

Issue #280, opened 2021-10-27
Formatting Attributes Applied to Entire Columns not Carrying Over to Saved Workbook 


mendy-friedman:
**Describe the bug**
If you read in a workbook with openxlsx, and it includes formatting that is applied to an entire column/row (for example, the entire column is shaded a certain color, or all columns from column X to the end of the workbook are hidden), this formatting is stripped upon saving the workbook out again.

**To Reproduce**
Initial file: 
[test.xlsx](https://github.com/ycphs/openxlsx/files/7427111/test.xlsx)

I read this in and saved it out again, and got this file.  As you can see, all columns are unhidden and A:F on the Sheet4 tab are no longer gray all the way down the sheet. (Hidden tabs also unhide, but this is already called out in a separate open issue: https://github.com/ycphs/openxlsx/issues/279#issue-1036796632).
[test-output - 4.2.4.xlsx](https://github.com/ycphs/openxlsx/files/7427113/test-output.-.4.2.4.xlsx)

**Expected behavior**
The formatting should be preserved upon loading and saving the workbook.

JanMarvin:
Confirmed. We are somehow losing a style. 

Yours looks like this:

```XML
<fills count="4">
  <fill>
    <patternFill patternType="none"/>
  </fill>
  <fill>
    <patternFill patternType="gray125"/>
  </fill>
  <fill>
    <patternFill patternType="solid">
    <fgColor theme="6" tint="0.59999389629810485"/>
    <bgColor indexed="64"/>
    </patternFill>
  </fill>
  <fill>
    <patternFill patternType="solid">
    <fgColor theme="6"/>
    <bgColor indexed="64"/>
    </patternFill>
  </fill>
</fills>
```

Ours looks like this:
```XML
<fills count="3">
  <fill>
    <patternFill patternType="none"/>
  </fill>
  <fill>
    <patternFill patternType="gray125"/>
  </fill>
  <fill>
    <patternFill patternType="solid">
    <fgColor theme="6"/>
    <bgColor indexed="64"/>
    </patternFill>
  </fill>
</fills>
```

Edit: it might be something else. We're ignoring quite a few xml nodes from styles. Most likely because nobody had the time or need to understand them.

JanMarvin:
I dug into the code and assume I've found the issue. 

The `loadworksheets()` function checks if the worksheet contains data.

[...]

PR #283, opened 2021-10-28
import a few styles nodes

JanMarvin:
These were found tackling #280


JanMarvin:
This is based on #282

Issue #231, opened 2021-07-26
xlsx file created with openxlsx can't be opened in SPSS


deschen1:
I saved some R data as an xlsx file with openxlsx in R. However, when trying to open this file in SPSS I get an error:

![grafik](https://user-images.githubusercontent.com/25662795/126949500-a158d61f-5fe8-4d22-9e84-9286ec70c98d.png)

When exporting the data with e.g. the `writexl` package, I don't get such an error. This happens regularly, i.e. many files exported with openxlsx can't be opened in SPSS anymore so my guess is that openxlsx somehow deviates from the standard meta structure of an xlsx file?

I'm also attaching the problematic Excel file (DELETED, see updated files here: https://github.com/ycphs/openxlsx/issues/231#issuecomment-886518071)

BTW: opening the problematic file in Excel directly and save it under a new name fixes the issue, i.e. the file can be opened in SPSS then.

JanMarvin:
Hi @deschen1 , when you mentioned that they can't be opened anymore I assume that previously they used to work? Could you please attach your fixed Excel too, it will be useful if someone wants to see where the issue is. Unless it's something pretty obvious, I'm not sure how we should debug the SPSS xlsx importer (if you're not depending on the xlsx files, you can use other packages to directly create spss binary files, I've written such a package myself).

Since there are multiple ways to create files with openxlsx, we'd need an minimal reproduce able example. Some R code that creates the file above.

If you have some time, it would be beneficial, if you could help narrowing down which commit has caused the issue. A good command line tool for this is git bisect, otherwise downloading a few commits, building the package, running the minimal example and opening the file in SPSS should make the deal. (It's somewhat tedious, but such is software development and if you're not doing it, someone else with access to SPSS has to, or fix it by mistake sometime in the future.)

deschen1:
@JanMarvin: I tested with some new code and the error is completely reproducible with whatever content I have in may data frame. So here's a small reproducible example:

```
df <- data.frame(uuid    = "asjdhbsadkbask",
                 var1_en = "This is just a test")

export_df <- openxlsx::createWorkbook()
openxlsx::addWorksheet(export_df, sheetName = "Sheet 1")
openxlsx::writeData(export_df, sheet = "Sheet 1", x = df)
openxlsx::saveWorkbook(export_df, file.choose())
```

As for your remaining message, I have to admit that I don't understand much about it. I'm rather an enduser and not a software developer, so I post things here, but otherwise don't know anything about github or commits.

[...]

PR #286, opened 2021-11-03
To avoid "Specified part does not exist in the package" error when op…

### LINKED ISSUE #285: tag in [Content_Types] refers to empty drawing, triggering error in OpenXml
**Describe the bug**
When there are no drawing objects, the xlsx file contains no /xl/drawings folder but the [Content_Types].xml still contains a tag that refers to a file in that folder. The file will still open OK in Excel, but it will throw an error message in OpenXls: "Specified part does not exist in the package." Files created directly by Excel that don't have drawings do not have the offending tag in [Content_Types].xml

**To Reproduce**
```
library(openxlsx)
wb <- createWorkbook()
addWorksheet(wb, 'Sheet1')
saveWorkbook(wb, 'drawing.in.content.types.but.no.drawing.file.in.xlsx.xlsx')
# Creates  xlsx file with tag in [Content_Types].xml that refers to PartName /xl/drawings/drawing1.xml
# However, there is no /xl/drawings folder and therefore no /xl/drawings/drawing1.xml file
# Excel opens the file OK, but OpenXls throws an error with the following error message:
# "Specified part does not exist in the package"
# 
# If the [Content_Types].xml file is edited to delete the offending tag then the file opens fine in OpenXml
# and also still opens fine in Excel
```

**Expected behavior**
A file that will import into Openxls error-free

**Screenshots**

**Example files**
The R code above produces [drawing.in.content.types.but.no.drawing.file.in.xlsx.xlsx](https://github.com/ycphs/openxlsx/files/7465331/drawing.in.content.types.but.no.drawing.file.in.xlsx.xlsx), which triggers the error in OpenXml.
If that file is opened in Excel, where an immediate "File | Save As" produces [file.saved.by.excel.xlsx](https://github.com/ycphs/openxlsx/files/7465332/file.saved.by.excel.xlsx) then that new file does *not* have the offending tag in [Content_Types].xml

**Additional context**
I have forked-and-cloned openxlsx and pushed a fix to acobolew/openxlsx in branch remove-empty-drawing-from-Content_Types . I will submit a PR with the fix

acobolew:
Fixes #285

### DIFF
--- a/R/WorkbookClass.R
+++ b/R/WorkbookClass.R
@@ -974,6 +974,10 @@ Workbook$methods(
         )
     }
 
+    for (i in seq_along(worksheets))
+      if (length(drawings[[i]])==0)
+        ct <- ct[!grepl(sprintf("drawing%s.xml", i), ct)]
+    
     ## write [Content_type]
     write_file(
       head = '<Types xmlns="http://schemas.openxmlformats.org/package/2006/content-types">',


JanMarvin:
Hi @acobolew , thanks for the issue and the pull request. Your proposal is kind of a hack, the "better" solution would be to fix the content creation file and only include the drawing folder if it's required. Though that might create a lot more headaches, therefore I'd merge this pull request 👍 
PD: Iirc there is some other open issue regarding the issue, I'll see if I can find it.

JanMarvin:
This is fix 1 from https://github.com/ycphs/openxlsx/issues/231#issuecomment-907809082

JanMarvin (approved):

JanMarvin:
If you want to, please add a sentence to the news file, otherwise looking good!

PR #289, opened 2021-11-07
makes `options()` setting consistent

jmbarbone:
This simplifies setting standard options like `scipen` and `Out.Dec` used throughout the package.  There shouldn't be any changes, just a little cleanup.


R/wrappers.R
```diff
@@ -92,14 +91,8 @@ createWorkbook <- function(creator = ifelse(.Platform$OS.type == "windows", Sys.
 #' saveWorkbook(wb, file = "saveWorkbookExample.xlsx", overwrite = TRUE)
 #' }
 saveWorkbook <- function(wb, file, overwrite = FALSE, returnValue = FALSE) {
-  od <- getOption("OutDec")
-  options("OutDec" = ".")
-  on.exit(expr = options("OutDec" = od), add = TRUE)
-
-  ## increase scipen to avoid writing in scientific
-  sci_pen <- getOption("scipen")
-  options("scipen" = 10000)
```
JanMarvin, line 101: after the change this defaults to scipen 200, is that enough? I'm unaware why this specific value was chosen, but we should make sure that we're not regressing in the xlsx output.

R/wrappers.R
```diff
@@ -604,13 +592,8 @@ renameWorksheet <- function(wb, sheet, newName) {
 convertFromExcelRef <- function(col) {
 
   ## increase scipen to avoid writing in scientific
-  exSciPen <- getOption("scipen")
-  od <- getOption("OutDec")
-  options("scipen" = 10000)
```
JanMarvin, line 609: same here

JanMarvin (approved):
I'm not sure about these two lines, though I trust your code

R/wrappers.R
jmbarbone: Some arbitrarily large number is assigned to try to preserve digits during conversions.  I think `200` is used in mostly places but anything large enough should probably be fine.

I guess `200` is twice as large as `100` -- that's the extent to which I thought about why.

JanMarvin: Thanks, perfectly fine for me. Though I'd appreciate @ycphs reviewing this PR too.

JanMarvin:
Thanks, merged it!

PR #291, opened 2021-11-08
Improve detectDates. closes #288

### LINKED ISSUE #288: `read.xlsx(detectDates = TRUE)` failing
Reporting again here as we still have this error

``` r
openxlsx::read.xlsx("C:/Users/jmbar/Downloads/bad.xlsx", detectDates = TRUE)
#> Error in read_workbook(cols_in = cell_cols, rows_in = cell_rows, v = v, : basic_string::substr: __pos (which is 8) > this->size() (which is 7)
```

<sup>Created on 2021-11-06 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

**Original**
 

> I can confirm the issue is still there. See the file attached, opening it with `detectDates=TRUE` raises an error:
> ```
> Error reading date:
> 44489.4
> row: 1
> col: 1
> Error in read_workbook(cols_in = cell_cols, rows_in = cell_rows, v = v,  : 
>   basic_string::substr: __pos (which is 8) > this->size() (which is 7)
> ```
> [bad.xlsx](https://github.com/awalker89/openxlsx/files/7491399/bad.xlsx)
> 
> _Originally posted by @aushev in https://github.com/awalker89/openxlsx/issues/249#issuecomment-962523430_

JanMarvin:
Hackish attempt to fix #288

This PR checks if there is a numeric value, at a position, where we already expect a string of this type "yyyy-mm-dd".  For whatever kind of reason, we end up here, with a numeric value in #288. Most likely due to the date being a POSIXct. Therefore most likely the fix is at the wrong position and should be somewhere else. For this minimal test file it should not matter, though with larger files we might waste a lot of time testing the conversion.


src/write_file.cpp
```diff
@@ -225,16 +226,21 @@ SEXP buildMatrixMixed(CharacterVector v,
...
         }else{
-          // dt_str = as<std::string>(m(ri,i));
           dt_str = m(ri,i);
+
           try{
-            datetmp[ri] = Rcpp::Date(atoi(dt_str.substr(5,2).c_str()), atoi(dt_str.substr(8,2).c_str()), atoi(dt_str.substr(0,4).c_str()) );
-          }catch(...) {
+            // Some values are incorrectly detected as dates. This regex determines if they are numerics.
+            // If so, they are converted to Dates. 
+            Rcpp::Function cTD("convertToDate");
```
JanMarvin: should be moved into the if condition or  out of the loop to avoid being called unnecessarily

cha-petersumm:
I still have issue, with a file that was simply written by openxlsx and then read back again (so Excel has never been near it).

write.xlsx(CVAD_list, "CVAD list.xlsx")
CVAD_list2 <- read.xlsx("CVAD list.xlsx")
CVAD_list3 <- read.xlsx("CVAD list.xlsx", detectDates = T)

Error message from the third line above is as follows:

Error reading date:
44696.5
row: 59
col: 6
Error: basic_string::substr: __pos (which is 8) > this->size() (which is 7)

There's nothing obviously odd about cell F59.

I'm happy to create an example that I can post if it would be useful.

packageVersion("openxlsx")
[1] ‘4.2.5.2’

JanMarvin:
Hi @cha-petersumm , you can have a look at `openxlsx2`. The changes here went into the development branch, but never to CRAN.

cha-petersumm:
> Hi @cha-petersumm , you can have a look at `openxlsx2`. The changes here went into the development branch, but never to CRAN.

Thanks.  That does appear to load from CRAN now, and it does fix the problem with dates.  Great work!

As per https://cran.r-project.org/web/packages/openxlsx2/index.html.

PR #297, opened 2021-11-19
Improve grouping with groupRows. closes #294

### LINKED ISSUE #294: Multiple groupRows?
It seems you can only group one set of rows per worksheet with this function. I want to create 3 separate, non-nested groups in the iris data set by Species. Can this currently be done?

JanMarvin:
allow different grouping options for `groupRows()`. Needs  testing due possible breaking changes to `write_file_2.cpp`.

### DIFF
--- a/R/wrappers.R
+++ b/R/wrappers.R
@@ -4533,6 +4533,8 @@ ungroupColumns <- function(wb, sheet, cols) {
 #' @examples 
 #' wb <- createWorkbook()
 #' addWorksheet(wb, 'Sheet1')
+#' addWorksheet(wb, 'Sheet2')
+#' 
 #' writeData(wb, "Sheet1", iris) 
 #' writeData(wb, "Sheet2", iris)
 #' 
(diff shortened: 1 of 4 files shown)


JanMarvin:
If this passes the checks (did not check prio to pushing) this needs some testing. I have made changes to the write function and got mad and did not try to fully understand the old code. Also I was not entirely sure how to handle grouping, because I do not really use it and therefore might have implemented changes that are unwanted/unneeded. Checked basic functionality with LO and MS 365.

bfreed3:
I figured out what I needed to with the old groupRows function. For the iris df, to group on the first instance of each Species in the xlsx file, the following code works for what I need.

groupRows(wb, sheet=1, c(2:51,53:101,103:151)) 

One just needs to play around with it a bit, and understand the shifting of rows going from the df to xlsx, and how Excel groups things on its side.  

Thanks, Burgess

JanMarvin:
Thanks for the follow up. I kinda figured the same when creating this pull request. You can't have different grouped rows in the same outline that are attached to each other. Excel will automatically combine both. To avoid this, you need to skip at least one row. Though this pull request allows having multiple nested groups and it fixes the previous warning message.
If you want, you could add a pull request to further improve the documentation. Which is quite short for this function and lead to the issue.

JanMarvin:
darn, I got lost with open pull requests (293 != 294) anyone got a copy of this branch lying around? :smile: 
Edit: fixed and restored. `git reflog` ftw!

PR #302, opened 2021-11-22
fix warning in extLinksXML. closes #293

### LINKED ISSUE #293: Invalid format string in loadWorkbook
As of R 4.1.0 the following bit throws a warning if `extLinksXML` is not empty since the format string does not contain any conversion specifications. I _think_ its last part should read `Target="externalLinks/externalLink%s.xml"` but I haven't tested that or thought through the logic of what's being done exactly.

https://github.com/ycphs/openxlsx/blob/08de207e0c069d75429ecec10f5d6d1b92e9cd46/R/loadWorkbook.R#L449-L452

JanMarvin:


R/loadWorkbook.R
```diff
@@ -452,7 +452,7 @@ loadWorkbook <- function(file, xlsxFile = NULL, isUnzipped = FALSE) {
     )
 
     wb$workbook.xml.rels <- c(wb$workbook.xml.rels, sprintf(
-      '<Relationship Id="rId4" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/externalLink" Target="externalLinks/externalLink1.xml"/>',
+      '<Relationship Id="rId4" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/externalLink" Target="externalLinks/externalLink%s.xml"/>',
```
JanMarvin, line 455: I'm wondering if `rId4` is identical for all or if we're supposed to loop over `rId +`1? Is there any way to test this?

JanMarvin:
I have no clue how to test this, can anybody have a look?

PR #310, opened 2021-12-27
when duplicating the style objects, assign the new sheet name

JanMarvin:
I have a workbook, where style 165 is already assigned after loading. When cloning, a new (duplicated) entry for `numFmtId=165` is assigned. Looking at the code, I stumbled over this.


JanMarvin:
Thinko. sheetName is already the new worksheet

PR #316, opened 2022-01-11
Do not allow `NA` as column name

### LINKED ISSUE #292: Corrupt Excel file when column name is `NA`
**Describe the bug**
When a dataframe's column name is NA, `openxlsx` writes a corrupt Excel file (according to Microsoft Excel 2016):

**To Reproduce**
```r
df <- data.frame("A", "B")
colnames(df) <- c(1, NA)

wb <- openxlsx::createWorkbook()
openxlsx::addWorksheet(wb, "sheet")
openxlsx::writeData(wb, 1, df)
openxlsx::saveWorkbook(wb, "temp.xlsx", overwrite = TRUE)
```

**Screenshots**
image

jmbarbone:
Resolves #292 

@JanMarvin do you know how we can tell if the `NA` is converted to `"NA"` in the sheet data?

The change is in `writeData` so it should have an effect in `buildWorkbook()`.  Just couldn't get my head around what `writeData()` and `Workbook$writeData()` were doing.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -22,6 +22,7 @@
 * `writeData()` calls `force(x)` to evaluate the object before options are set ([#264](https://github.com/ycphs/openxlsx/issues/264))
 * `createComment()` now correctly handles `integers` in `width` and `height` ([#275](https://github.com/ycphs/openxlsx/issues/275))
 * `setStyles()` accepts `halign="justify"` ([#305](https://github.com/ycphs/openxlsx/issues/305))
+* `data.frame`s with `NA` column names are converted to `"NA"` when saving ([#292](https://github.com/ycphs/openxlsx/issues/292), [#316](https://github.com/ycphs/openxlsx/issues/316)
 
 # openxlsx 4.2.4
 
--- a/R/writeData.R
+++ b/R/writeData.R
@@ -180,6 +180,8 @@ writeData <- function(
 ) {
 
   x <- force(x)
+  # transforms NA to "NA" for bad names
+  names(x) <- paste(names(x))
   
   op <- get_set_options()
   on.exit(options(op), add = TRUE)
--- a/tests/testthat/test-writeData.R
+++ b/tests/testthat/test-writeData.R
@@ -35,3 +35,17 @@ test_that("writeData() forces evaluation of x (#264)", {
   options(op)
   file.remove(wbfile)
 })
+
+test_that("colnames with NA are appropriately handled [292]", {
+  x <- data.frame(a = 1, b = 2)
+  colnames(x) <- c("a", NA)
+  wbfile <- temp_xlsx()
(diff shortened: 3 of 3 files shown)


JanMarvin:
You just want to know, if the string is imported as "NA"? If written, is is converted into a `sharedString` (`<sst><si /></sst>` in openxml). Therefore you could check if `wb$sharedString` contains `"<si><t xml:space=\"preserve\">NA</t></si>"`.

The position in the shared string is a 0 index, matching the value of `v + 1` in `wb$worksheets[[1]]$sheet_data$v` where the type is `1` in `wb$worksheets[[1]]$sheet_data$t`.

I know that feeling in `writeData` and friends. The construction of the shared string and the lists are deeply embedded in `openxlsx` and make it near impossible to really implement things like `inlineStr`.

JanMarvin:
You could look into `make.names()` to create unique names, it is used in the package at various places, otherwise you might end up with multiple "NA"s which brings other problems.

jmbarbone:
> You could look into `make.names()` to create unique names, it is used in the package at various places, otherwise you might end up with multiple "NA"s which brings other problems.

Considered that.  Not sure if it's that's to best.  There's no requirement for data in sheets to be unique, so the `"NA"` seemed like a less invasive fix.  `make.names()` is used for the reading, which makes it difficult to test this (thought I could make the workbook and then read it to make sure it was okay).  Technically you can still have a `data.frame()` in R without unique names, it's just a bad idea:

``` r
x <- data.frame(x = 1)
x <- cbind(x, x)
x
#>   x x
#> 1 1 1
str(x)
#> 'data.frame':    1 obs. of  2 variables:
#>  $ x: num 1
#>  $ x: num 1
```

<sup>Created on 2022-01-11 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

I've seen plenty of workbooks with repeated column names.  Maybe something a little nicer?

``` r
x <- c(NA, "a", "b", NA, "b", "a", NA)

make.names(x)
#> [1] "NA." "a"   "b"   "NA." "b"   "a"   "NA."
make.names(x, unique = TRUE)
#> [1] "NA."   "a"     "b"     "NA..1" "b.1"   "a.1"   "NA..2"

clean_na_names <- function(x) {
  w <- which(is.na(x))
  for (i in w) {
    x[w] <- paste0("NA..", w)
  }
  x
}

clean_na_names(x)
#> [1] "NA..1" "a"     "b"     "NA..4" "b"     "a"     "NA..7"
```

<sup>Created on 2022-01-11 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

What I'm worried about is disharmony between reading and writing though.  Maybe warnings/controls for reading a workbook as a `data.frame` with duplicated column names?

PR #322, opened 2022-01-23
Update writeFormula.Rd

christianbraathen:
Clarifying that arguments must be comma-separated.

### DIFF
--- a/man/writeFormula.Rd
+++ b/man/writeFormula.Rd
@@ -19,7 +19,7 @@ writeFormula(
 
 \item{sheet}{The worksheet to write to. Can be the worksheet index or name.}
 
-\item{x}{A character vector.}
+\item{x}{A character vector. Arguments must be comma-separated.}
 
 \item{startCol}{A vector specifying the starting column to write to.}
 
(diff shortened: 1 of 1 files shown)


JanMarvin:
Hi @christianbraathen, thanks for the pull request.The change should be made in the roxygen2 tag in the R file. Once it is added, you should run roxygen2. This will generate the man page (the one you have edited). If you want to, more details in the description or examples section would be welcome too.

christianbraathen:
Hi @JanMarvin ! It's the first time I contribute to a package. Would you mind giving me a quick description with code on how to reproduce the docs properly? Then I'll create another PR.

JanMarvin:
Sure. Just add it here:
https://github.com/ycphs/openxlsx/blob/110fb4fec0a461a2367dabf8f2c5f203b99f977d/R/writeData.R#L433-L518

Afterwards you run roxygen2 (`roxygen2::roxygenize()` (check if that updates the man page after installation) or ctrl+shift+D in Rstudio). This ensures that the man pages are updated with code changes.

PR #328, opened 2022-02-16
Fix grouped rows and columns (Issue #138)

### LINKED ISSUE #138: Error when saving Workbook containing grouped rows
**Describe the bug**
Loading an existing Excel file that contains some rows that are grouped and saving it leads to the following error message:.

> Error in write_worksheet_xml_2(prior = prior, post = post, sheet_data = ws$sheet_data,  : 
>   CHAR() can only be applied to a 'CHARSXP', not a 'NULL'
> 

**To Reproduce**
File containing 1 sheet with small data and some rows are grouped:
[has_grouped_rows.xlsx](https://github.com/ycphs/openxlsx/files/5879375/has_grouped_rows.xlsx)

**Fails**:  Load existing workbook that contains grouped rows and save it:
```
library(openxlsx)
wb = loadWorkbook("has_grouped_rows.xlsx")
saveWorkbook(wb, file="has_grouped_rows_out.xlsx")
```

**Works**: ungroup rows prior to saving workbook
```
wb = loadWorkbook("has_grouped_rows.xlsx")
ungroupRows(wb, 1, 1:20)`
saveWorkbook(wb, file="has_grouped_rows_out.xlsx")
```

**Works**: ungroup rows and group again prior to saving workbook
```
wb = loadWorkbook("has_grouped_rows.xlsx")
ungroupRows(wb, 1, 1:20)
groupRows(wb, 1, 2:11)
saveWorkbook(wb, file="has_grouped_rows_out.xlsx")
```

**Desktop:**
tested on two different systems with identical behavior:

System 1:
 - Windows 10
 - Excel Version: Microsoft Excel for Windows 365 MSO (16.0.13328.20476) 64-Bit
 - R Version: R4.0.2
 - openxlsx: 4.2.3

System 2:
 - Windows 10
 - Excel Version: Microsoft Excel for Windows 2016 MSO (16.0.13530.20418) 64-Bit
 - R Version: R4.0.3
 - openxlsx: 4.2.3

**Additional context**
This problem does not apply to grouped columns, workbooks containing grouped columns can be saved without problems.

kainhofer:
This series of patches fixes several aspects of row/column grouping:
-) Correctly preserve grouping and hidden flag upon loading for all cases
-) Fix creating nested groups and ungrouping rows/columns
-) Make sure all grouped rows / columns are written to the XML file in the correct order
-) Simplify the logic to create the XML row elements for grouped rows in the xlsx file.
-) Add unit test case for all scenarios

This fixes #138 for good and also fixes several other open (unreported) issues with grouped rows/columns.


kainhofer:
Does anyone know, why the link check always  times out after 6 hours? Apparently, this should not be caused by the pull request, but seems a larger issue with the check here at github. This timeout holds back this pull request, as the checks appear as failed...

JanMarvin:
@kainhofer , I've already reported the `lintr` issue upstream. We'll see what happens. It should not hold this PR back. Though I'd like @ycphs to review this one. I don't feel like adding people to DESCRIPTION is my call to make. I have nothing against it, but it's the package owners call to make.

JanMarvin:
The `lintr` issue is solved. Upstream found an unmatched open `{` in a vignette that caused the issue. A pull request was merged in master (as it didn't effect code and will stop pull requests opened against master, from timing out, until the issue is fixed in `lintr`). I've merged it to `development` so a merge/rebase from there should solve the issue here too.

kainhofer:
> I don't feel like adding people to DESCRIPTION is my call to make. I have nothing against it, but it's the package owners call to make.

I might add that I didn't add myself for this patch alone. Rather, I have also contributed other features (like the cloneWorksheet functionality), but apparently never bothered to add myself to the contributors list back then.

ycphs:
@kainhofer,  thank you for the contribution.

PR #331, opened 2022-02-23
Bug #330: writeData with a date creates a new number format style, which potentially needs a unique number format ID

### LINKED ISSUE #330: writeData with a date messes up existing cell styles (numFmt)
**Describe the bug**
I have an excel file with several cells with custom number format ("+0;-0;0").

Loading and saving that workbook with openxlsx works fine. However, as soon as writeData with a date object is used to insert a date anywhere in the file, the cells with the custom number format are turned into date cells. (See the screenshots below).

Digging deeper into the XML, the reason is that date cells are inserted as numeric cells with a custom number format. Unfortunately, that new custom number format for date cells uses the same numFmtId as my custom number format (styles.xml inside the .xlsx file):
image

**To Reproduce**
R code to reproduce (with the attached sample file):
```{R}
library(openxlsx)
fmt = openxlsx::loadWorkbook("openxlsx_numberformat_date.xlsx")

# unmodified file can be saved and the format is preserved:
openxlsx::saveWorkbook(fmt, "openxlsx_numberformat_date_out.working.xlsx", overwrite = TRUE) 
openXL("openxlsx_numberformat_date_out.working.xlsx")

# Inserting a date object ANYWHERE in the file messes up the format:
openxlsx::writeData(fmt, 1, x = as.Date("2021-01-01"), xy = c(1,1))
openxlsx::saveWorkbook(fmt, "openxlsx_numberformat_date_out.broken.xlsx", overwrite = TRUE)
openXL("openxlsx_numberformat_date_out.broken.xlsx")
```

**Expected behavior**
The existing cell number format should be preserved.

**Screenshots**
Original file (in LibreOffice): 
image

After openxlsx inserts a date in some other cell:
image

**Example files**
[openxlsx_numberformat_date.xlsx](https://github.com/ycphs/openxlsx/files/8121035/openxlsx_numberformat_date.xlsx)

**Additional context**
As mentioned above, inserting a date object with writeData will add a custom number format to the cells. 
However, the function classStyles (file helperFunctions.R) creates that style by a call to createStyle(numFmt = "date").
Unfortunately, the createStyle function (file wrappers.R) hard-codes the numFmtId to 165:
```{R}
  if (numFmt != "general") {
    if (numFmt %in% validNumFmt) {
      style$numFmt <- numFmtMapping[[numFmt[[1]]]]
    } else {
      style$numFmt <- list("numFmtId" = 165, formatCode = numFmt) ## Custom numFmt
    }
  }
```
This hard-coded ID causes duplicate IDs for both the existing number format and the new date format. 

The proper solution would be to detect the highest used numFmtId in the workbook and make sure that the ID used is higher than any of the used IDs.

kainhofer:
Make sure that writeData with a date creates a style that has no collisions of its numFmtId with existing styles. By default, a style with numFmdId=165 (hard-coded!) is created by createStyle. When inserting that style into a workbook, we need to assign it a new, unique numFmtId if the workbook already has some custom number formats.
The explicitly-called addStyle function checked for this case, but the implicitly-called classStyles function did not.

Fixes #330.

### DIFF
--- a/R/helperFunctions.R
+++ b/R/helperFunctions.R
@@ -132,12 +132,12 @@ classStyles <- function(wb, sheet, startRow, startCol, colNames, nRow, colClasse
   names(colClasses) <- NULL
   
   # For custom number formats, ensure unique IDs (extract the current maximum and add 1 for each new format)
-  maxnumFmtId <- max(c(sapply(wb$styleObjects, function(i) {
+  maxnumFmtId <- max(unlist(sapply(wb$styleObjects, function(i) {
     as.integer(
       max(c(i$style$numFmt$numFmtId, 0))
     )
-  }), 165))
-  
+  })), 165)
+
   
   if ("hyperlink" %in% allColClasses) {
     
--- a/R/helperFunctions.R
+++ b/R/helperFunctions.R
@@ -131,6 +131,14 @@ classStyles <- function(wb, sheet, startRow, startCol, colNames, nRow, colClasse
   newStylesElements <- NULL
   names(colClasses) <- NULL
   
+  # For custom number formats, ensure unique IDs (extract the current maximum and add 1 for each new format)
+  maxnumFmtId <- max(c(sapply(wb$styleObjects, function(i) {
+    as.integer(
+      max(c(i$style$numFmt$numFmtId, 0))
+    )
+  }), 165))
+  
+  
   if ("hyperlink" %in% allColClasses) {
     
     ## style hyperlinks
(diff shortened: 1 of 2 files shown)


JanMarvin:
If you could fix the tests we can merge this. It's rather small, a test would be an additional benefit as well as checks for dups in numfmtid, but these are optional. So long nobody has complained and the package is around for a long time :-)
Some entry in NEWS.md would be nice.

JanMarvin (changes requested):
Unittests fail.

mgross-ebner:
> If you could fix the tests we can merge this. It's rather small, a test would be an additional benefit as well as checks for dups in numfmtid, but these are optional. So long nobody has complained and the package is around for a long time :-) Some entry in NEWS.md would be nice.

Dear Jan, just for the sake of completeness, let me mention that we are affected by the problem and are eagerly waiting for a patch. 

Thank you for the super package and your efforts!

Best regards
Markus

ycphs:
As soon as the unit tests are fixed, we can merge the PR. I totally understand the need you have. If you could provide a fix for the unit tests, we can merge the PR.

I am looking forward for your contribution, @mgross-ebner

kainhofer:
@ycphs I'm aware of the failing unit tests (and they indicate real problems with the patch, which need to be fixed). 
Unfortunately, I'm currently really short on time, so I can't promise when I'll be able to take a look and fix the issues with the PR.

kainhofer:
Just fixed the issue with the unit tests. Apparently sapply(list(....), function(x) ...) will return a vector, UNLESS the list is empty, in which case an empty list (and not an empty vector) is returned... That's quite inconsistent behavior of sapply, but can be worked around by a call to unlist to always force the results of sapply to a vector...

PR #336, opened 2022-03-12
Get file mode from output directory and not from tempdir. fix #333

### LINKED ISSUE #333: Files should be written according to default ACLs
Expected Behavior

Files should be written using the system acls
Actual Behavior

For example it works properly with writexl package,
But it doesn't using openxlsx.
Need to set Sys.umask("0007") in .Rprofile in my case to have proper acls
Steps to Reproduce the Problem

(please attach an example xlsx file if possible)

    Mount an NFS directory with specific acls

    Make a script to write an xls file to this directory

    Check the file rights

sessionInfo()

    Version of openxlsx: `4.2.5.9000`
    Version of R: `4.1.1`

I don't know if this is linked to the NFS mount but I prefer to mention this specificity.
What seems odd to me is that the behaviour is different from writexl and openxlsx.
The former one seems right ...

Here's a file with right acls written with writexl package :

library(writexl)
cars <- mtcars[1:6, ]
write_xlsx(x = cars, path = "test_writexl.xlsx", col_names = TRUE)

Here's another script with wrong acls with openxlsx package :

library("openxlsx")
wb <- createWorkbook()
cars <- mtcars[1:6, ]
addWorksheet(wb, "Cars")
writeData(wb, "Cars", cars, startCol = 2, startRow = 3, rowNames = TRUE)
saveWorkbook(wb, "test_write_openxlsx.xlsx", overwrite = TRUE)

Here are the results :

-rw-------  1 myuser test-projet 6875 Mar  2 19:22 test_write_openxlsx.xlsx
-rw-rw-rw-  1 myuser test-projet 5438 Mar  2 19:23 test_writexl.xlsx

And here are the ACLs for my current directory on the NFS server :

user::rwx
group::---
group:test-projet:rwx
other::---
default:user::rwx
default:group::---
default:group:test-projet:rwx
default:other::rwx

And here are the ACLs on the mounted directory (same directory but client side)

user::rwx
group::rwx
other::

JanMarvin:
Hi @finalspy, please check if this solves your issue. I am not sure how this affects Windows, but this are base R functions, maybe we are lucky.

If creating the file in `/tmp` is the issue:

1) We write some empty file of the identical name to the output directory.
2) We get the correct access rights.
3) Once the zip file is created, we change the output access rights.

### DIFF
--- a/R/wrappers.R
+++ b/R/wrappers.R
@@ -112,8 +112,12 @@ saveWorkbook <- function(wb, file, overwrite = FALSE, returnValue = FALSE) {
 
   xlsx_file <- wb$saveWorkbook()
 
-
-  file.create(file)
+  # file did not exist, create it and overwrite
+  if (!file.exists(file)) {
+    file.create(file)
+    overwrite <- TRUE
+  }
+  
   def_mode <- file.info(file)$mode
   result <- file.copy(from = xlsx_file, to = file, overwrite = overwrite)
   Sys.chmod(file, def_mode)
--- a/R/wrappers.R
+++ b/R/wrappers.R
@@ -112,16 +112,7 @@ saveWorkbook <- function(wb, file, overwrite = FALSE, returnValue = FALSE) {
 
   xlsx_file <- wb$saveWorkbook()
 
-  # # file did not exist, create it and overwrite
-  # if (!file.exists(file)) {
-  #   file.create(file)
-  #   overwrite <- TRUE
-  # }
-  #
-  # def_mode <- file.info(file)$mode
   result <- file.copy(from = xlsx_file, to = file, overwrite = overwrite, copy.mode = FALSE)
-  # Sys.chmod(file, def_mode, FALSE)
-
 
   ## delete temporary dir
   unlink(dirname(xlsx_file), force = TRUE, recursive = TRUE)
--- a/R/wrappers.R
+++ b/R/wrappers.R
@@ -112,7 +112,12 @@ saveWorkbook <- function(wb, file, overwrite = FALSE, returnValue = FALSE) {
 
   xlsx_file <- wb$saveWorkbook()
 
+
+  file.create(file)
+  def_mode <- file.info(file)$mode
   result <- file.copy(from = xlsx_file, to = file, overwrite = overwrite)
+  Sys.chmod(file, def_mode)
+
 
   ## delete temporary dir
   unlink(dirname(xlsx_file), force = TRUE, recursive = TRUE)
--- a/R/wrappers.R
+++ b/R/wrappers.R
@@ -112,15 +112,15 @@ saveWorkbook <- function(wb, file, overwrite = FALSE, returnValue = FALSE) {
 
   xlsx_file <- wb$saveWorkbook()
 
-  # file did not exist, create it and overwrite
-  if (!file.exists(file)) {
-    file.create(file)
-    overwrite <- TRUE
-  }
-  
-  def_mode <- file.info(file)$mode
-  result <- file.copy(from = xlsx_file, to = file, overwrite = overwrite)
-  Sys.chmod(file, def_mode)
+  # # file did not exist, create it and overwrite
+  # if (!file.exists(file)) {
+  #   file.create(file)
+  #   overwrite <- TRUE
+  # }
+  #
+  # def_mode <- file.info(file)$mode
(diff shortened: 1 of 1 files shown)


JanMarvin:
Fixed my previous thinko, hopefully unittests will like this one better.

JanMarvin:
The one Mac test is broken, the rest is fine. Still someone has to check with Windows nevertheless. Anyhow, this should set the access rights accordingly to previous versions of the file one wants to replace or to the default access rights inside a folder.

finalspy:
I made several tests and it partialy solves the problem.

I updated my [Docker test project] (https://github.com/finalspy/openxlsx333)
And tag 4 different versions and made available 4 docker images on [DockerHub](https://hub.docker.com/repository/docker/finalspy/openxlsx333) that reproduces the 4 tests

Here are the versions and their differences :

 - 0.1.0 uses the latest openxlsx release and default umask 0022
 - 0.1.1 uses the latest openslsx release and forces umask 0002
 - 0.2.0 uses your fix branch and default Umask 0022
 - 0.2.1 uses your fix branch and forces umask 0002

Note : I used a `reference.file` which is created by a simple touch, to illustrate diffrences between generated files fom openxlsx or other pachage.

The results ares the following : 

## Version 0.1.0
Initial version, shows a difference in file rights between openxlsx qnd reference file.

    ```
    -rw-r--r--+ 1 my_user my_user 6864 Mar 13 22:26 openxlsx.xlsx
    -rw-rw----+ 1 my_user my_user    0 Mar 13 22:26 reference.file
    -rw-rw----+ 1 my_user my_user 5438 Mar 13 22:26 writexl.xlsx
    ```

## Version 0.1.1
Same as previous  but tests changes on umask to see if it has some kind of influence.
It shows that umask doesn't changes write permissions of files (at least openxlsx written one as it should). This means that as previously stated, openxlsx rights are inherited from somewhere else (not hte current directory nor the umask).   

    ```
    -rw-r--r--+ 1 my_user my_user 6864 Mar 13 22:26 openxlsx.xlsx
    -rw-rw----+ 1 my_user my_user    0 Mar 13 22:26 reference.file
    -rw-rw----+ 1 my_user my_user 5438 Mar 13 22:26 writexl.xlsx
    ```

## Version 0.2.0
Using the fix branch, rights seems more conform, but write permission is still missing compared to `reference.file` for instance.

    ```
    -rw-r-----+ 1 my_user my_user 6864 Mar 13 22:26 openxlsx.xlsx
    -rw-rw----+ 1 my_user my_user    0 Mar 13 22:26 reference.file
    -rw-rw----+ 1 my_user my_user 5438 Mar 13 22:26 writexl.xlsx
    ```

## Version 0.2.1
With this version all the rights are the same, but only after changing the default umask. We qre on the good way but something is still wrong and escapes to my understanding ... 

    ```
    -rw-rw----+ 1 my_user my_user 6864 Mar 13 22:26 openxlsx.xlsx
    -rw-rw----+ 1 my_user my_user    0 Mar 13 22:26 reference.file
    -rw-rw----+ 1 my_user my_user 5439 Mar 13 22:26 writexl.xlsx
    ```

 wdyt ?

JanMarvin:
Thanks for the tests! We have at least two other options to try.

1) `file.copy()` has `copy.mode` as option, missed it, when I created my initial pull request. This should use the file mode in the "to" directory.
2) If that doesn't help, we can try to set `use.umask=FALSE` in `Sys.chmod()`, it is in the commented part of the code above.

Let me know if either of this helps (preferably the latest commit), otherwise I have to spin up the docker image myself.

finalspy:
Pushed a new [test image tagged 0.2.0_20220314_01](https://hub.docker.com/repository/docker/finalspy/openxlsx333/tags?page=1&ordering=last_updated),
[source unchanged from 0.2.0](https://github.com/finalspy/openxlsx333), only forced rebuild with --no-cache to take your latest commit into account.

And everything seems OK to me, without having to force a specific umask as long as parent directories default ACLs are set :

    -rw-rw----+ 1 my_user my_user 6864 Mar 14 07:07 openxlsx.xlsx
    -rw-rw----+ 1 my_user my_user    0 Mar 14 07:07 reference.file
    -rw-rw----+ 1 my_user my_user 5438 Mar 14 07:07 writexl.xlsx

Great Job !

JanMarvin:
Hooray! I'll cleanup the mess and merge it to development. From there it will go into master and the next release. Thank you for testing!

PR #341, opened 2022-04-10
Feature Request: Merge repeated cells by one or more variables

darentsai:
When data have one or more categorical variables, it's common to sort them and merge repeated cells together when creating an excel file. The argument `mergeCols` added in `writeData()` controls which columns need to be merged. Its order  indicates how your data is grouped and sorted. An example using `esoph` data is demonstrated as follows:

```r
df <- with(esoph, esoph[order(agegp, alcgp, tobgp), ])[1:30, ]
wb <- createWorkbook()
addWorksheet(wb, "esoph")
writeData(wb, "esoph", df, borders = "all", mergeCols = 1:3)
addStyle(wb, "esoph", createStyle(halign = "center", valign = "center"),
         rows = 1:(nrow(df)+1), cols = 1:ncol(df), gridExpand = TRUE, stack = TRUE)
saveWorkbook(wb, "writeDataExample.xlsx", overwrite = TRUE)
```
example


JanMarvin:
Hi @darentsai , lets begin with "it's common": I have never seen someone doing something like this and I've seen people doing many strange things. Therefore I have do disagree that it's common ;)

Otherwise it looks good. Just a suggestion, maybe the sorting should also be done via writeData? So that we keep the input and output in sync. Are there any drawbacks that I am overlooking?

Final remarks:
 * a test for this would be awesome (something like your code snippet and for example some detection of how many merged cells the workbook contains) and 
 * a rerun of roxygen.

darentsai:
@JanMarvin Thanks for your kind reply. My statements may make some misunderstanding. If the data written to an excel file is used for further analysis or modeling, this feature is not practical. But, if the data have been grouped, summarised by some categorical variables and going to be presented in a report, this feature can make the appearance of table much tidier.(I update a new example in my previous comment using another dataset.)

As for the sorting process, I also thought about doing it in `writeData`. In my opinion, a method that prints data to an external file should not change the input structure (e.g. order). Hence, I presume users have arranged data as they want before writing it out. `mergeCols` only works on how the data is rendered to excel.

JanMarvin:
Alright, I agree on the sorting part. Give @ycphs or me a ping when to review (still like to see a test, even if it's just a tiny `expect_silent(write...)`).

JanMarvin:
Hi @darentsai , what is the state of this? Ready and just missing a test or did you find some larger issue that is holding this pull request back?

darentsai:
@JanMarvin Thanks for the reminder. I have tried several kinds of data and adjusted different sets of arguments. I don't find anything wrong so far. I think it's ready to be used but lacks for a test. I'm sorry that I'm not familiar with the test mechanism of package development, so I haven't figured out a proper test example to examine this new feature.

Issue #313, opened 2022-01-11
Excel fail to open the output created from openxlsx with slicer


radiance88:
**Describe the bug**
I created a simple excel file "Slicer_test.xlsx", with some random data, a pivot table and a slicer.
Nothing wrong when using R openxlsx. I simply use loadWorkbook to load it and then saveWorkbook to save it as another file.
Then when I open the new file it says 'We found a problem with some content in the file. Do you want us to try to recover".
If I click "Yes", it says "The workbook cannot be opened or repaired because it is corrupt"
**To Reproduce**

wb <- loadWorkbook("Slicer_test.xlsx")
saveWorkbook(wb,"Slicer_test_saved.xlsx")

**Expected behavior**
I expect Excel should be able to open the output file with no problem.

**Screenshots**
![image](https://user-images.githubusercontent.com/9069384/148888524-194e131c-2411-4dec-84cf-e466c48d644a.png)
![image](https://user-images.githubusercontent.com/9069384/148888725-1b7e64af-1d80-45a1-85bc-d098d760807a.png)

**Example files**
If applicable, add example files to help explain y
[Slicer_test.xlsx](https://github.com/ycphs/openxlsx/files/7844027/Slicer_test.xlsx)
our problem (`openxlsx` and/or `Excel` created files)

JanMarvin:
Thanks, confirmed. In the created file `xl/workbook.xml` contains multiple entries of `extLst`. That is not supposed to happen and Excel most likely does not like that. I guess I imported that because we were losing some extLst entries, now we have them duplicated. Oh boy

JanMarvin:
A fix for now:
```R
# import your Slicer_test.xlsx
wb <- openxlsx::loadWorkbook("~/gh_issue_313.xlsx")
# keep only the last extLst (the openxlsx created one in this case)
wb$workbook$extLst <- wb$workbook$extLst[2]
# works in ms365
openxlsx::saveWorkbook(wb, "test.xlsx", overwrite = TRUE)
```

JanMarvin:
For some reason `openxlsx` decided, that it wants some `extLst` (most likely due to the slicer found on a sheet), though we failed to check if there are extLst elements we could merge with or even overwrite. Appending seems like the worst idea in this case as it leads to broken files for Excel. If bitten by this: they work fine with LibreOffice.

Issue #370, opened 2022-07-19
Opening a wb object that was loaded from xlsx causes content problems


jzadra:
**Describe the bug**
If you create a wb from scratch and open it in excel, it works fine.  If you instead load a wb from an existin xlsx file and then try to open in in excel, it causes an excel alert regarding problematic conent.

**To Reproduce**
``` r
library(tidyverse)
library(openxlsx)

wb <- loadWorkbook(file = system.file("extdata", "loadExample.xlsx", package = "openxlsx"))

openXL(wb)
```
![image](https://user-images.githubusercontent.com/25590353/179849223-e6b014bc-cc15-495a-b82d-86bd52e365fa.png)
 
Clicking "yes" yields:

![image](https://user-images.githubusercontent.com/25590353/179849349-3dbc5d6d-d4d4-4f0a-b5d6-ded85cd44fea.png)

**Expected behavior**
The workbook should open, and should be identical to the original workbook.

─ Session info ────────────────────────────────────────────────────────────────────────────────────────
 setting  value
 version  R version 4.2.0 (2022-04-22)
 os       macOS Monterey 12.4
 system   x86_64, darwin17.0
 ui       RStudio
 language (EN)
 collate  en_US.UTF-8
 ctype    en_US.UTF-8
 tz       America/Denver
 date     2022-07-19
 rstudio  2022.02.3+492 Prairie Trillium (desktop)
 pandoc   2.17.1.1 @ /Applications/RStudio.app/Contents/MacOS/quarto/bin/ (via rmarkdown)

─ Packages ────────────────────────────────────────────────────────────────────────────────────────────
 ! package     * version date (UTC) lib source
   assertthat    0.2.1   2019-03-21 [2] CRAN (R 4.2.0)
   backports     1.4.1   2021-12-13 [2] CRAN (R 4.2.0)
   broom         0.8.0   2022-04-13 [2] CRAN (R 4.2.0)
   callr         3.7.0   2021-04-20 [2] CRAN (R 4.2.0)
   cellranger    1.1.0   2016-07-27 [2] CRAN (R 4.2.0)
   cli           3.3.0   2022-04-25 [2] CRAN (R 4.2.0)
   clipr         0.8.0   2022-02-22 [2] CRAN (R 4.2.0)
   colorspace    2.0-3   2022-02-21 [2] CRAN (R 4.2.0)
 P crayon        1.5.1   2022-03-26 [?] CRAN (R 4.2.0)
[...]

PR #371, opened 2022-07-20
remove reading extlst

JanMarvin:
this reverts https://github.com/ycphs/openxlsx/commit/312ea6c22c5f49da25c5bc4e9e82d085d2dd5f2c


JanMarvin:
might fix #313 #370

JanMarvin:
The issue is: with the reverted commit we read the correct extLst node. Though if the workbook contains a slider, we create a custom extLst node for the slider. Since we are unable to efficiently merge XMLs, we append two extLsts which a certain spreadsheet software is unable to recover from.

PR #373, opened 2022-07-26
Add auto row_heights feature

David-Rattray:
Hi @JanMarvin,

Here is the fork with the updates from @DavidBreuer 's code.

As I mentioned in the issue (#367 ) thread, I've encountered an error with 
`Error in write_worksheet_xml_2(prior = prior, post = post, sheet_data = ws$sheet_data, : Not compatible with STRSXP: [type=NULL].`

I think it stems from having non-null row heights in the workbook when I go to write, and the write function
`write_worksheet_xml_2 <- function(prior, post, sheet_data, row_heights_ = NULL, outline_levels_ = NULL, R_fileName = "output") {
    .Call(`_openxlsx_write_worksheet_xml_2`, prior, post, sheet_data, row_heights_, outline_levels_, R_fileName)
}`

However it is just a hunch

PR #375, opened 2022-07-30
Add auto row_heights feature

JanMarvin:
Fixed code by @DavidBreuer pushed by @David-Rattray in #373.


David-Rattray:
Hi @JanMarvin been busy/unwell for awhile, but aiming to get back on this.
As you mentioned in the closed pull request, I'll look to get feature tests and backward testing sorted out, and along the way maybe clean things up a smidgeon. I'm going to just read https://r-pkgs.org/testing-basics.html and hope that gets me far enough. Let me know if you have any recommendations for learning material.

JanMarvin:
Hi @David-Rattray , welcome back and I hope you're better now! I'd write a test that expects a certain row height. Though I do not know if it works as expected. Therefore visual testing is required in the first place.

Btw I closed your initial PR to quick. I can push to your PR, but you can't push to this one. Either we reopen your PR or you open a new one.

David-Rattray:
@JanMarvin Yeah so would the smartest thing for me be to make a new fork from the development branch to get the changes you implemented, then use that to add the feature tests and then create a new pull request when I'm finished? Or would there be a better branch to grab the code changes from?

Edit: Just looked again, probably should clone/fork from auto_resize?

David-Rattray:
@JanMarvin to save you replying I just cloned auto_resize branch and am working away. Feature tests on hold for the moment. Was looking at the code closer, it appears all the auto does is add the extra height parameter to the base height parameter. Going to look closer at what setColumnWidths does to auto adjust and see if I can implement something similar for row heights.

JanMarvin:
But wasn't that expected? I assumed that it calculates which row height is required and this value is adjusted in the workbook.

David-Rattray:
Yeah I jumped the gun a bit when I was reading the functions in wrappers.R didn't realize the autoHeights bit was under helper functions. Seemed like it was just adding a static extra height parameter not a calculated one and the test output has a slightly too short row height for the sample text.

PR #383, opened 2022-09-23
Preserve window size and position, also get and set

michael-dewar:
This PR will preserve the size and position of the window when you open the xlsx in Excel.  There are also functions to set and get directly.


JanMarvin:
Hi @michael-dewar , looking good. Just minor lintr issues and I'd like to see a test for this as well.

michael-dewar:
I fixed the linter issues for the code I changed.  I mostly avoided changing the other sections because I wasn't actually sure if  the & vs && change made a difference compared to what was intended.

I also added tests.

JanMarvin:
Thanks, the other changes should be in the `development` branch

Issue #403, opened 2023-02-11
`saveWorkbook` produces invalid XML when manipulating libreoffice based XLSX files


retostauffer:
**The issue:**

I am using `loadWorkbook()`, `writeData()` and `saveWorkbook()` to
modify an existing XLSX file, originally created with **libreoffice** on linux.

The first time (manipulating the XLSX file once with `openxlsx`) works well, the
second time the XLSX gets corrupted. `openxlsx::read.xlsx` is still able to read
the data, however, libreoffice only shows empty sheets, and `readxl::read_excel`
refused to read the data (`Error: unexpected end of data`).

I may have found the problem (unclosed XML tag) as shown in the last part of this issue. **Note** that this error does not occur when I start with an XLSX file created (saved) by/via MS Excel.

**System information:**

* Tested with `4.2.2` on Windows and Ubuntu linux
* `openxlsx` version `4.2.5.2' (current CRAN release)
* `openxlsx` version `4.2.5.9000` (main branch of this repos)

**Minimal:**

I have put together a small minimal which uses a file [test_libreoffice.xlsx](https://github.com/ycphs/openxlsx/files/10714147/test_libreoffice.xlsx),
a simple XLSX file with two sheets. This XLSX file has been created using
_LibreOffice 7.3.7.2 30(Build:2)_ on an _Ubuntu Ubuntu 22.04.1 LTS_.

The _R_ script below does the following:

1. Tests if `"test_libreoffice.xlsx" is readable
2. Manipulates `"test_libreoffice.xlsx"`, stores modified version as `"test1.xlsx"`
3. Tests readability of `"test1.xlsx"`
4. Manipulates `"test1.xlsx"`, stores modified version as `"test2.xlsx"`
5. Tests readability of `"test2.xlsx"`

Testing readability (`test_read()`) tries to read the first sheet with both,
`openxlsx` and `readxl` (testing). On both, my linux system as well as on
my Windows 10 it fails in (5) where `readxl::read_excel()` is no longer to
read the data. At this point, libreoffice will also only show empty sheets (no data).

```
library("readxl")
library("openxlsx")

# Reading sheet 1 with both openxlsx as well as readxl
test_read <- function(file) {
    cat("Testing readability of file \"", file, "\"\n")
    print(read.xlsx(file, sheet = 1))
    print(dim(read_excel(file, sheet = 1)))
}

# Small function which overwrites the content of the first sheet
modify_file <- function(src, dst) {
    # New (static) data set
[...]

PR #404, opened 2023-02-11
Properly closing </sheetPr> when writing `xl/worksheets/sheet*.xml`

retostauffer:
Dear dev team,

writing issue #403 seems to have sorted my thoughts and I was able to track down the problem (what I think the problem is). This pull request fixes the problem with the minimal (see #403) and does not break the tests.

Thanks for all your work and all best,
R


JanMarvin:
Hi @retostauffer , I fear that the pull request is based on an incorrect assumption. The duplicated sheetPr tag needs to be removed, not closed. Otherwise Excel will not be able to open the file. At least that's what I do in openxlsx2 and that's tested with Excel.

PR #433, opened 2023-06-08
Improve offline testing

nanxstats:
This pull request improves the handling of offline scenarios in tests that require an internet connection.

Specifically, it adds a `skip_if_offline()` call to these tests, which will prevent them from running when there is no internet connection detected.

This change improves the resilience of the unit tests and make it easier for people to run tests in different environments such as air gapped ones.

### DIFF
--- a/tests/testthat/test-read_sources.R
+++ b/tests/testthat/test-read_sources.R
@@ -8,6 +8,7 @@ context("Read Sources")
 
 
 test_that("read.xlsx from different sources", {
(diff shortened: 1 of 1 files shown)


nanxstats:
Hey @ycphs, could you take a quick look at this PR when you get a chance? I think it improves one aspect of the tests but not sure if there are other considerations when they were first written. Your feedback would be highly valuable. Thanks!

PR #468, opened 2024-04-29
Add `pair_rc()` to improve row/column pair lookups

### LINKED ISSUE #467: Speed up row/column pair lookups
**Is your feature request related to a problem? Please describe.**
There are are few places in the codebase where row and column indices are combined with `paste()`. This is done in order to check a vector of pairs against another vector of pairs. For large datasets, this produces two large character vectors. 

**Describe the solution you'd like**
We could "hash" the combinations in some way (e.g. a simple pairing function would be ideal). Better yet would be a list-wise %in% operator, like the one the [collapse package](https://search.r-project.org/CRAN/refmans/collapse/html/fmatch.html) provides.

StevenHibble-Concurrence:
Closes #467.

I considered [Szudzik's Elegant Pairing Function](http://szudzik.com/ElegantPairing.pdf), but it quickly exceeds `.Machine$integer.max`. Instead, we create a double vector by taking the row position as the whole number and the column position as the decimal number. This assumes that float precision is better than 1e-7 and that the spreadsheet won't have more than 10M-1 columns.

```
(1, 1)     ->   1.0000001
(999, 999) -> 999.0000999
```

### DIFF
--- a/R/WorkbookClass.R
+++ b/R/WorkbookClass.R
@@ -3362,12 +3362,9 @@ Workbook$methods(
         sId <-
           .self$updateStyles(this.sty) ## this creates the XML for styles.XML
 
-        cells_to_style <- stri_join(x$rows, x$cols, sep = ",")
-        existing_cells <-
-          stri_join(worksheets[[sheet]]$sheet_data$rows,
-            worksheets[[sheet]]$sheet_data$cols,
-            sep = ","
-          )
+        cells_to_style <- pair_rc(x$rows, x$cols)
+        existing_cells <- pair_rc(worksheets[[sheet]]$sheet_data$rows,
+                                  worksheets[[sheet]]$sheet_data$cols)
 
         ## In here we create any style_ids that don't yet exist in sheet_data
         worksheets[[sheet]]$sheet_data$style_id[existing_cells %in% cells_to_style] <<-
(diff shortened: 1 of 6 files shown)


StevenHibble-Concurrence:
I forgot to add some metrics. I took timing and total memory allocation using `bench::mark()`. Peak RAM usage is from `peakRAM::peakRAM()`.

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<body link="#467886" vlink="#96607D">

workbook   size | metric | paste() | pair_rc() | change
-- | -- | -- | -- | --
medium | time (seconds) | 7.2 | 6.4 | -11%
medium | mem_alloc (MB) | 399.8 | 355.7 | -11%
medium | peak_mem (MB) | 283.7 | 263.3 | -7%
large | time (seconds) | 46.4 | 33.1 | -29%
large | mem_alloc (MB) | 3696.6 | 3184.6 | -14%
large | peak_mem (MB) | 449.5 | 395.2 | -12%
</body>
</html>

JanMarvin:
Thanks!

PR #472, opened 2024-06-03
Fix rdevel segfaults

JanMarvin:
https://cran.r-project.org/web/checks/check_results_openxlsx.html


JanMarvin:
Oh I recreated the data files with the data creation file. Otherwise the tests were failing, not sure why they were removed in the first place.

PR #473, opened 2024-06-03
Fix rdevel segfaults2

JanMarvin:
Hi @ycphs , your merge was a little to quick for me :)

I wanted to at least make the CI pass. There are a bunch of `lintr` and remaining other issues that I wont tackle. I was just curious to see if the segfault can be fixed and doing my part to keep the package on CRAN.

Fingers crossed that it will build now run without errors.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -2,11 +2,13 @@
 # Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
 on:
   push:
-    branches: [main, master, development]
+    branches: [main, master]
   pull_request:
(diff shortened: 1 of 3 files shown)


JanMarvin:
Alright R-CMD-check run was a success! Adding curl as a dependency for three occurrences of a testthat function might be something to consider.

1) You might want to have a second look at my update of the github actions script in the previous commit. After a short night and a long day, I was losing it a bit and simply replaced everything with the current example (similar for this commit, but ...).

2) A long time ago, I started to merge a few commits into the development branch. Maybe if you go over the package the next time, you can resurrect a few.

PR #482, opened 2024-07-24
silence -Wformat warning

JanMarvin:
CRAN has a few warnings currently only visible with older clang on Mac. `-Wformat` on gcc and clang on Linux does not complain.

```c++
read_workbook.cpp:555:52: warning: format specifies type 'unsigned short' but the argument has type 'int' [-Wformat]
read_workbook.cpp:565:54: warning: format specifies type 'unsigned short' but the argument has type 'int' [-Wformat]
read_workbook.cpp:626:50: warning: format specifies type 'unsigned short' but the argument has type 'int' [-Wformat]
```

### DIFF
--- a/src/read_workbook.cpp
+++ b/src/read_workbook.cpp
@@ -536,12 +536,12 @@ SEXP read_workbook(IntegerVector cols_in,
   CharacterVector col_names(nCols);
   IntegerVector removeFlag;
   int pos = 0;
+  char name[7];
   
   // If we are told col_names exist take the first row and fill any gaps with X.i
   if(hasColNames){
     
     int row_1 = rows[0];
-    char name[6];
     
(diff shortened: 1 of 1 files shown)


JanMarvin:
Cannot reproduce on Mac either. Maybe a spurious warning in clang14?

```R
using C++ compiler: ‘Apple clang version 15.0.0 (clang-1500.3.9.4)’
using SDK: ‘MacOSX14.4.sdk’
```

src/read_workbook.cpp
```diff
@@ -536,12 +536,12 @@ SEXP read_workbook(IntegerVector cols_in,
   CharacterVector col_names(nCols);
   IntegerVector removeFlag;
   int pos = 0;
+  char name[7];
```
JanMarvin: Probably can be removed now

PR #466, opened 2024-04-19
Preserve window size and position, also get and set

michael-dewar:
This PR will preserve the size and position of the window when you open the xlsx in Excel. There are also functions to set and get directly.

I had earlier created a PR #383, but it got hopelessly out-of-date with the changes to the package.  I closed that old PR and created this new one.  My hope is that this will allow a cleaner merge, assuming people find this PR useful.

PR #487, opened 2024-08-01
Preserve window size and position, also get and set (Clean Version)

michael-dewar:
This PR will preserve the size and position of the window when you open the xlsx in Excel. There are also functions to set and get directly. 

This PR replaces earlier PRs #466 and #383.  Those got out of date and I was unsuccessful at properly rebasing them.

This code is lean in that it only touches the bare minimum of the original code.  It also adds one file with two new functions, a file with tests for those functions, and related documentation.

The outputted `xlsx` files open cleanly on popular spreadsheet software on Mac and Windows, and web versions of spreadsheet software.


JanMarvin:
Thanks, looks good!

PR #494, opened 2024-08-27
Fix resaving workbook after resetting col widths

### LINKED ISSUE #493: Issue when setting col widths and saving workbook multiple times
**Describe the bug**
There's a bug in the Workbook object method `setColWidths()` (not the exported function by the same name) that causes `saveWorkbook()` to return an error when called after the user has set column widths for a range of columns (e.g. 1:2), saved the workbook, then set column widths for a new range that is inclusive of the previous one (e.g. 1:5).

The error that is returned is:
```
Error in sprintf("<col min=\"%s\" max=\"%s\" width=\"%s\" hidden=\"%s\" customWidth=\"1\"/>",  : 
  arguments cannot be recycled to the same length
```

**To Reproduce**
The following code demonstrates the issue:
```
# Write some data to a workbook object.
wb <- createWorkbook()
addWorksheet(wb, "iris") 
writeData(wb, "iris", iris)

# Set column widths and save the workbook.
setColWidths(wb, "iris", cols = 1:2, 12)
tf <- tempfile(fileext = ".xlsx")
saveWorkbook(wb, file = tf, overwrite = TRUE)

# Set column widths again for a different range inclusive of the previous one and save the workbook.
setColWidths(wb, "iris", cols = 1:5, 15)
saveWorkbook(wb, file = tf, overwrite = TRUE)
```

**Additional context**
Here `cols` is subset to the columns that have not previously been set, but `widths` and `hidden` are not:
https://github.com/ycphs/openxlsx/blob/2bf325a950344c399dff7a0c352adf416c092b40/R/workbook_column_widths.R#L157

This means that these three objects are no longer the same length when they are passed to `sprintf()` here:
https://github.com/ycphs/openxlsx/blob/2bf325a950344c399dff7a0c352adf416c092b40/R/workbook_column_widths.R#L162

jistria:
This PR resolves the issue described in #493 (error returned when resaving workbook after resetting col widths) by ensuring that the `cols`, `widths` and `hidden` objects inside the Workbook method `setColWidths()` have the same length before being passed to `sprintf()`.

It's success is demonstrated by the following code:
```
# Write some data to a workbook object.
wb <- createWorkbook()
addWorksheet(wb, "iris") 
writeData(wb, "iris", iris)

# Set column widths and save the workbook.
setColWidths(wb, "iris", cols = 1:2, 12)
tf <- tempfile(fileext = ".xlsx")
saveWorkbook(wb, file = tf, overwrite = TRUE)

# Set column widths again for a different range inclusive of the previous one and save the workbook.
setColWidths(wb, "iris", cols = 1:5, 15)
saveWorkbook(wb, file = tf, overwrite = TRUE)
```

Fixes #493

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,7 @@
 
 * Fixed warning on `dataValidation(..., type = "list")` ([#342](https://github.com/ycphs/openxlsx/issues/342))
 * Added optional argument to `loadWorkbook` to decide if empty/blank cells should be converted to NA_character_ (the default) or left blank as is
+* `saveWorkbook()` now succeeds when called after the user has set column widths for a range of columns (e.g. 1:2), saved the workbook, then set column widths for a new range that is inclusive of the previous one (e.g. 1:5) ([#493](https://github.com/ycphs/openxlsx/issues/493)).
 
 ## Improvements
 
--- a/R/workbook_column_widths.R
+++ b/R/workbook_column_widths.R
@@ -154,7 +154,10 @@ Workbook$methods(setColWidths = function(sheet) {
       }
     }
 
-    cols <- cols[!cols %in% names(worksheets[[sheet]]$cols)]
+    remaining_cols <- !cols %in% names(worksheets[[sheet]]$cols)
+    cols <- cols[remaining_cols]
+    widths <- widths[remaining_cols]
+    hidden <- hidden[remaining_cols]
   }
 
   # Add remaining columns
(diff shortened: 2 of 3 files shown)


JanMarvin:
Thanks, could you please add a test for the fix?

jistria:
Thanks for reviewing this PR, @JanMarvin. I've added a test for the fix now.

JanMarvin:
Thanks!

PR #495, opened 2024-08-30
4.2.7

JanMarvin:


JanMarvin:
Thanks everyone involved. It took a while, but 4.2.7 is on its way to CRAN.

PR #505, opened 2024-10-01
Fix the docs for setLastModifiedBy and getCreators

PBfordev:
These two functions had their @title same as addCreator().


JanMarvin:
Thanks @PBfordev !

PR #515, opened 2024-12-06
fix sharedString deletion

### LINKED ISSUE #514: deleteDataColumn adds strings in wrong places
# Issue
Hi, this is a weird glitch that took some time to reproduce. Here is a sample excel file: 
[before.xlsx](https://github.com/user-attachments/files/17989495/before.xlsx)
image

Then I run this code to load the file, delete the 3rd column (C) and save to a new file:
```
library(openxlsx)

my_wb <- loadWorkbook("before.xlsx")

openxlsx::deleteDataColumn(my_wb, sheet =  1, col =  3)

saveWorkbook(my_wb, "example_after.xlsx", overwrite = T)
```

And the output looks like this:
image
You can see that in C2 there is the string "d" that should not be there. (The cell in the column named  "d" should be an institution name, as before the column removal)

## Extra context
It was hard to reproduce because when I changed the column names for my reprex, all of a sudden it worked without issue! The cell EXTRA originally had some other text. I changed it for privacy, which changed what string gets copied. Instead of "d", the cell C2 used to have content from A3 pasted into it. This changed after I manually edited the excel. Maybe it has something to do with the order in which cells were filled?

## Possible cause
I'm not certain what causes this issue. My best guess would be something with pointers to shared strings, but I don't understand the topic well enough to see the problem. If I had to guess, using the "keep" object in the following line seems weird to me because it seems not to take into account the deleting of strings in the previous lines.

https://github.com/ycphs/openxlsx/blob/69e00afa3780076c8d0eff21b9beb5c13f819b49/R/wrappers.R#L1929

Thank you for your hard work!

DavZim:
Fixes #514 

Also adds additional tests for the shared string usecase.

Makes the deleteDataColumn code a little bit easier to read and maintain.

Fixes another bug with sharedStrings across sheets. Now sharedStrings are fixed on a workbook-level and not sheet-level.

### DIFF
--- a/R/wrappers.R
+++ b/R/wrappers.R
@@ -1900,44 +1900,65 @@ deleteDataColumn <- function(wb, sheet, col) {
   }
 
   a <- wb$worksheets[[sheet]]$sheet_data
+  
+  # t: if a shared string is used or if the string is a value itself
+  # v: the shared string index or the string itself (if t == 0)
+  # in wb$sharedStrings we find the values of the shared strings by index + 1 (0 indexed!)
+  
 
   # check which elements to delete
   keep <- a$cols != col
   # if there is no column to delete, exit early
   if (all(keep)) return(invisible(0))
 
-  # delete cols in cols "col" move higher cols one down
+  # delete cols in cols "col", move higher cols one down
   a$cols <- as.integer(a$cols[keep] - 1 * (a$cols[keep] > col))
   a$rows <- a$rows[keep]
 
   # reduce the shared strings pointers if they are not used anymore
-  has_t <- !is.na(a$t) & a$t == 1
-  used_shared <- a$v[has_t] # a reference to all shared strings
-  keep_t <- keep[has_t] # these shared strings are kept
-  keep_t[is.na(keep_t)] <- FALSE
-  keep_shared <- used_shared[keep_t]
-  rem_shared <- setdiff(unique(used_shared[!keep_t]), unique(keep_shared))
-  for (v in rem_shared) {
-    to_reduce <- as.numeric(keep_shared) > as.numeric(v)
-    to_reduce[is.na(to_reduce)] <- FALSE
-    if (any(to_reduce))
-      keep_shared[to_reduce] <- as.character(as.numeric(keep_shared[to_reduce]) - 1)
-  }
-  used_shared[keep_t] <- keep_shared
-  a$v[has_t] <- used_shared
-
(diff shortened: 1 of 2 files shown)


R/wrappers.R
```diff
@@ -1900,45 +1900,69 @@ deleteDataColumn <- function(wb, sheet, col) {
...
-      keep_shared[to_reduce] <- as.character(as.numeric(keep_shared[to_reduce]) - 1)
-  }
-  used_shared[keep_t] <- keep_shared
-  a$v[has_t] <- used_shared
-
+  ss <- data.frame(old = numeric(0), new = numeric(0), string = character(0))
+  if (length(wb$sharedStrings) > 0) {
+    ss <- data.frame(
+      # the old index 0 indexed, as used in a$v
+      old = as.numeric(seq(length(wb$sharedStrings)) - 1),
```
JanMarvin, line 1923: Maybe `seq_along(wb$sharedStrings) - 1`

tests/testthat/test-wrappers.R
```diff
@@ -17,7 +17,7 @@ test_that("int2col and col2int", {
 test_that("deleteDataColumn basics", {
   wb <- createWorkbook()
   addWorksheet(wb, "tester")
-
+
```
JanMarvin: unintended whitespace changes in the test file

JanMarvin:
Hi @DavZim , thanks for taking care of the issue. Much appreciated!

Maybe one idea, you resize shared strings and subsequently have to update all references in all sheets. Is that required? Can't we just leave shared strings as is, or replace the removed strings with blanks or `NA_character_`, so that we do not have to adjust every shared strings entry?

DavZim:
Good question, I didn't think of this. Is it maybe worth adding it behind a function argument? Or do you think it's such an edge case that its better to keep shared strings anyways?

JanMarvin:
I doubt that it’s required to add new arguments for this, but I haven’t checked the impact yet. Aka what spreadsheet software does if some shared string entry isn’t used. Unless the shared strings are very long or very many, the impact on the file size should be negligible.

JanMarvin:
Hi, is this ready to merge? Aside from my usual minor nitpicking? I'm planing a release maybe at the end of the month.

DavZim:
I think this is ready to merge. It still cleans unused string references but as we have more tests, it should be good to merge

JanMarvin:
Thanks!

PR #539, opened 2026-07-12
Fix date format code length checks

sims1253:
Date-format detection used nchar(formatCodes > 3) instead of nchar(formatCodes) > 3. Consequently, the intended minimum-length check was ineffective and short date-like formats such as "m" could cause numeric cells to be incorrectly converted to dates.

Found with https://github.com/sims1253/ry

### DIFF
--- a/R/readWorkbook.R
+++ b/R/readWorkbook.R
@@ -483,7 +483,7 @@ read.xlsx.default <- function(
       
       ## this regex defines what "looks" like a date
       dateIds <- numFmtsIds[!grepl("[^mdyhsapAMP[:punct:] ]", formatCodes) &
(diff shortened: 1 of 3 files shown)


JanMarvin:
Thanks, looks indeed strange. Have you validated this with spreadsheet software? Like is numfmt “m” actually a month or is this a minute? Iirc the rules for this are a bit all over the place.

JanMarvin:
Thanks, its fine. Had a brief look! Do you need an urgent CRAN release for this?

sims1253:
I only discovered it when I ran my new typechecker on a bunch of r packages so there's no need from my side to get this released asap.

